PR #84, opened 2015-02-26
Add iOS 8 framework target

indragiek:
This is to facilitate using this library with tools like [Carthage](https://github.com/Carthage/Carthage), which are only compatible with framework targets.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,6 @@ before_install:
   - gem install xcpretty --no-rdoc --no-ri --no-document --quiet
 
 script:
-  - set -o pipefail && xcodebuild -project OHHTTPStubs/OHHTTPStubs.xcodeproj -scheme "OHHTTPStubs iOS" -sdk iphonesimulator -configuration Release ONLY_ACTIVE_ARCH=NO test | xcpretty -c
-  - set -o pipefail && xcodebuild -project OHHTTPStubs/OHHTTPStubs.xcodeproj -scheme "OHHTTPStubs Mac" -sdk iphonesimulator -configuration Release ONLY_ACTIVE_ARCH=NO test | xcpretty -c
+  - set -o pipefail && xcodebuild -project OHHTTPStubs/OHHTTPStubs.xcodeproj -scheme "OHHTTPStubs iOS StaticLib" -sdk iphonesimulator -configuration Release ONLY_ACTIVE_ARCH=NO test | xcpretty -c
+  - set -o pipefail && xcodebuild -project OHHTTPStubs/OHHTTPStubs.xcodeproj -scheme "OHHTTPStubs Mac Framework" -sdk macosx -configuration Release ONLY_ACTIVE_ARCH=NO test | xcpretty -c
+  - set -o pipefail && xcodebuild -project OHHTTPStubs/OHHTTPStubs.xcodeproj -scheme "OHHTTPStubs iOS Framework" -sdk iphonesimulator -configuration Release ONLY_ACTIVE_ARCH=NO test | xcpretty -c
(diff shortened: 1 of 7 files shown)


AliSoftware:
Nice, thanks! :+1:
I'll probably review this during the week-end.

I think it would be best to rename existing targets (and thus generated products) to make it more clear/explicit now, right?

* `OHHTTPStubs iOS` :arrow_right: `OHHTTPStubs iOS StaticLib`
* `OHHTTPStubs iOS Framework` :white_check_mark: 
* `OHHTTPStubs iOS Tests` :white_check_mark: 
* `OHHTTPStubs Mac` :arrow_right: `OHHTTPStubs Mac Framework`
* `OHHTTPStubs Mac Tests` :white_check_mark:

indragiek:
@AliSoftware Done!

indragiek:
@AliSoftware I added in one more fix that is irrelevant to the iOS framework, but fixes the installation directory on the Mac framework such that targets linking it will look for it in the right place. Without this, my test bundles fail to run with the following error message:

```
"The bundle “AlamofireRACExtensionsTests-Mac” couldn’t be loaded because it is damaged or missing necessary resources." (dlopen_preflight(/Users/Karunaratne/Library/Developer/Xcode/DerivedData/AlamofireRACExtensions-fwzqfpnjghfgdvgaqkdzjymwufxo/Build/Products/Debug/AlamofireRACExtensionsTests-Mac.xctest/Contents/MacOS/AlamofireRACExtensionsTests-Mac): Library not loaded: /Library/Frameworks/OHHTTPStubs.framework/Versions/A/OHHTTPStubs
  Referenced from: /Users/Karunaratne/Library/Developer/Xcode/DerivedData/AlamofireRACExtensions-fwzqfpnjghfgdvgaqkdzjymwufxo/Build/Products/Debug/AlamofireRACExtensionsTests-Mac.xctest/Contents/MacOS/AlamofireRACExtensionsTests-Mac
  Reason: image not found)
```

AliSoftware:
Cool, thanks a lot!

Will merge this soon :)

AliSoftware:
According to Travis, it seems that the `OHHTTPStubs iOS Framework` is not configured for Testing.

That can make sense not to run tests for this scheme on Travis (since we already ran the iOS Tests using `OHHTTPStubs iOS StaticLib` scheme), but I think we should:

* Configure the `OHHTTPStubs iOS Framework` for testing, and associate its test action to the `OHHTTPStubs iOS Tests` test target. That way if the scheme is selected in Xcode and the user hits `Cmd+U` then the tests for iOS are run
* But configure `.travis.yml` to only run the `build` (or `analyse`) action (and not the `test` action) on this `OHHTTPStubs iOS Framework` scheme, as the iOS Unit Tests will already have been tested by Travis when the `test` action would be run on the `OHHTTPStubs iOS StaticLib` scheme.

indragiek:
@AliSoftware Made those changes but it seems Travis is still failing because "iPad 2 isn't a valid run destination"? I'm not very familiar with Travis, any idea what's going on?

corinnekrych:
@indragiek running the command locally i get the ipad 2 pb too.
Adding to xcodebuild command
-destination 'platform=iOS Simulator,name=iPhone 6,OS=latest'
fixed it

AliSoftware:
In the meantime, I just opened a new issue #87 to push this further and add more travis destinations, especially because we should run  unit tests both for iOS7 and iOS8 to ensure there is no regression when there are changes in iOS8 internals like we have observed before.

_(Just to be clear: we should add the line described by @corinnekrych to `.travis.yml` for this PR to pass, and #87 will only be for after this has been merged)_

indragiek:
Looks like's it worked. Thanks for the tip @corinnekrych!

AliSoftware:
Thanks @indragiek for the terrific work.

As I was inspecting the code, I see that you added the exported variables `OHHTTPStubsFrameworkVersionNumber` and `OHHTTPStubsFrameworkVersionString`.
But the "Apple-Generic" versioning system (activated in the Build Settings on the `OHHTTPStubs iOS Framework` target) generates `OHHTTPStubsVersionNumber` and `OHHTTPStubsVersionString` variables (without the `Framework` in between)

We could fix that in the lines you added in `OHHTTPStubs.h` to remove the `Framework` part. But using the Apple-Generic versioning system means that I have to fill that version number each time `OHHTTPStubs` increase its version manually (and it can only be a double, not a 3-digits version number). That's error-inducing as I'm sure I'll forget to change that build setting each time I bump the pod version anyway.

Are you ok if I get rid of the Apple-Generic versioning system for the iOS Framework (as I did for the Mac Framework at that time) to be more consistant? Or did you have any specific reason to use it when you created the iOS framework target?

indragiek:
@AliSoftware No particular reason, feel free to remove it! Thanks for merging.

Issue #91, opened 2015-03-13
HTTP status code 300 will not be treated correctly by stubs


tarbrain:
We have noticed that HTTP status code 300 is not handled correctly.

300 represents "multiple choices" and involves user interaction, so according to the specification it should be treated similar to a 2xx response to give the user the chance to respond.

NSURLConnection behaves correctly by responding with ``` -connection:didReceiveResponse:``` while the stubs will treat it like a redirect which 300 is not.

AliSoftware:
Good catch, thanks!

Could you maybe make a PR with the fix? (I'll be quite busy the next few days and won't work on `OHHTTPStubs` for a while)

That's probably simply a `<` to change to `<=`, but the PR should also add a unit test that ensures it won't happen again ;)

Thx!

PR #92, opened 2015-03-14
Handle HTTP status code 300 differently than other redirect codes.

tarbrain:
Treats status code 300 'multiple choices' like a 2xx code to give the user agent a chance to select the redirect url.

Relates to issue #91


AliSoftware:
Perfect! Thanks a lot! :+1:

PR #95, opened 2015-04-01
Fix OHHTTPStubsProtocol threading

NSProgrammer:
Per Apple docs https://developer.apple.com/library/prerelease/ios/samplecode/CustomHTTPProtocol/Listings/Read_Me_About_CustomHTTPProtocol_txt.html :
```
In addition, an NSURLProtocol subclass is expected to call the various methods of the NSURLProtocolClient protocol from the client thread
```

So, instead of having callbacks happen on a background queue, OHHTTPStubsProtocol need to capture the executing runloop and execute callbacks from that runloop.  This is a requirement of NSURLProtocol.

-- Twitter uses OHHTTPStubs for certain unit tests and we have made this change on our local fork.  Providing the fix here for the benefit of all.


NSProgrammer:
There is a syntax bug, will resubmit

AliSoftware:
Thx! Note that there was no need to close that PR and recreate one. You can add new commits to your branch / fork before the PR gets merged and they will be added to the PR automatically by GitHub

Issue #97, opened 2015-04-03
Carthage support not working


MichaelMcGuire:
When adding OHHTTPStubs per the instructions in the README, the framework is not building.  Someone reported it on the Carthage project (Carthage/Carthage#405).  It looks like the Example project is found before the main project.  Not sure if you want to change the structure for Carthage support, but the README should probably at least be updated for now.

AliSoftware:
Thanks for reporting!

I am beginning to think that I'll drop Carthage support. Too much work as a lib maintainer, that's the 4th issue I had to fix just to workaround Carthage bugs so far in a short amount of time; each time I fix a thing for Carthage another issue arise… I don't have time for this.

I'll fix the README accordingly. Thanks!

YasKuraishi:
@AliSoftware - the fix was pretty trivial fix, just had to rename Example folder to Sample so Carthage could find the project file. PR waiting for your seal of approval... https://github.com/AliSoftware/OHHTTPStubs/pull/98

AliSoftware:
Now that I moved Example projects down a level (as I added a Swift demo project alongside the already existing ObjC demo project), Carthage should not use the Example project(s) anymore.

This modification has been done on `master` so feel free to test it. I'll release `4.0.0` later in the week-end once I finished removing some legacy code.

AliSoftware:
Closing due to lack of feedback. Should theoretically be fixed anyway (I don't use Carthage so won't test myself)

MichaelMcGuire:
Sorry for the lack of feedback!  I just converted over to using Carthage from CocoaPods on the latest in master (4.0.1) and everything is working great.  Thanks!

AliSoftware:
Thanks for letting me know :)

PR #98, opened 2015-04-08
Renamed `Example` folder to `Sample` to fix `Carthage` dependency manage...

YasKuraishi:
Fixed: https://github.com/AliSoftware/OHHTTPStubs/issues/97


AliSoftware:
This will probably break `pod try` support.

angrauel:
Please don't drop Carthage support. A lot of people use it instead of cocoapods

Goles:
I must +1 the petition of not dropping Carthage support :+1:

AliSoftware:
Be reassured, I won't drop it… but I won't corrupt/amend my project just to fix/workaround Carthage bugs either (and #97 — which this PR was a _workaround_ for — is indeed a Carthage issue, not mine).

rogernolan:
:+1: for that view. I think Carthage should add the flexibility to specify a target.

PR #103, opened 2015-05-20
Fix `OHResourceBundle` name mismatch between header and implementation

tibr:
`OHResourceBundle` did not work anymore because the function definition in the header did not match the name in the implementation.


AliSoftware:
Nice catch, thanks for the PR (and the additional unit test!) :+1:

AliSoftware:
`4.0.2` is out :)

tibr:
Thanks a lot for the quick release :+1:

PR #108, opened 2015-07-30
Stub network requests easily with Mocktail files/folders

JinlianWang:
This is an effort to support stubbing the network requests using files in the format of Mocktail(see more at: https://github.com/square/objc-mocktail). Mocktail is a popular format of network mocking which contains every piece of information in order to mock/stub a network request. By supporting this format, first, we will be able to allow Mocktail users to convert to OHHTTPStatus; Secondly, we will be able to add other functionalities to OHHTTPStatus, for example, recording traffic in the format of Mocktail and later on use it for stubbing.

### DIFF
--- a/OHHTTPStubs/Sources/Mocktail/OHHTTPStubs+Mocktail.h
+++ b/OHHTTPStubs/Sources/Mocktail/OHHTTPStubs+Mocktail.h
@@ -29,13 +29,14 @@ extern NSString* const MocktailErrorDomain;
  * This method will split the HTTP method Regex, the absolute URL Regex, the headers, the HTTP status code and
  * response body, and use them to add a stub.
  *
- * @param fileName The name of the `"*.tail"` file (with extension) in the Mocktail format.
+ * @param fileName The name of the mocktail file (without extension of '.tail') in the Mocktail format.
+ * @param bundleOrNil The bundle in which the mocktail file is located. If `nil`, the `[NSBundle bundleForClass:self.class]` will be used.
  * @param error An out value that returns any error encountered during stubbing. Returns an NSError object if any error; otherwise returns nil.
  *
  * @return a stub descriptor that uniquely identifies the stub and can be later used to remove it with
  * `removeStub:`.
  */
-+(id<OHHTTPStubsDescriptor>)stubRequestsUsingMocktailNamed:(NSString *)fileName error:(NSError **)error;
++(id<OHHTTPStubsDescriptor>)stubRequestsUsingMocktailNamed:(NSString *)fileName inBundle:(nullable NSBundle*)bundleOrNil error:(NSError **)error;
 
 /**
  * Add a stub given a file URL in the format of Mocktail as defined at https://github.com/square/objc-mocktail.
@@ -57,12 +58,13 @@ extern NSString* const MocktailErrorDomain;
  * This method will retrieve all the files under the folder; for each file with surfix of ".tail", it will split the HTTP method Regex, the absolute URL Regex, the headers, the HTTP status code and response body, and use them to add a stub.
  *
  * @param path The name of the folder containing files in the Mocktail format.
+ * @param bundleOrNil The bundle in which the path is located. If `nil`, the `[NSBundle bundleForClass:self.class]` will be used.
  * @param error An out value that returns any error encountered during stubbing. Returns an NSError object if any error; otherwise returns nil.
  *
  * @return an array of stub descriptor that uniquely identifies the stub and can be later used to remove it with
  * `removeStub:`.
  */
-+(NSArray *)stubRequestsUsingMocktailsAtPath:(NSString *)path error:(NSError **)error;
++(NSArray *)stubRequestsUsingMocktailsAtPath:(NSString *)path inBundle:(nullable NSBundle*)bundleOrNil error:(NSError **)error;
 
 @end
 
--- a/OHHTTPStubs/Sources/Mocktail/OHHTTPStubs+Mocktail.m
+++ b/OHHTTPStubs/Sources/Mocktail/OHHTTPStubs+Mocktail.m
@@ -12,10 +12,10 @@ NSString* const MocktailErrorDomain = @"Mocktail";
 @implementation OHHTTPStubs (Mocktail)
 
 
-+(NSArray *)stubRequestsUsingMocktailsAtPath:(NSString *)path error:(NSError **)error
++(NSArray *)stubRequestsUsingMocktailsAtPath:(NSString *)path inBundle:(nullable NSBundle*)bundleOrNil error:(NSError **)error
 {
-    NSString *dirPath = OHPathForFile(path, [self class]);
-    if (!dirPath)
+    NSURL *dirURL = [bundleOrNil?:[NSBundle bundleForClass:self.class] URLForResource:path withExtension:nil];
+    if (!dirURL)
     {
         if (error)
         {
@@ -25,17 +25,9 @@ NSString* const MocktailErrorDomain = @"Mocktail";
     }
     
     // Make sure path points to a directory
-    BOOL isDir = NO, exists = NO;
-    NSFileManager *fileManager = [NSFileManager defaultManager];
-    exists = [fileManager fileExistsAtPath:dirPath isDirectory:&isDir];
-    if (!exists)
-    {
-        if (error)
-        {
-            *error = [NSError errorWithDomain:MocktailErrorDomain code:OHHTTPStubsMocktailErrorPathDoesNotExist userInfo:@{NSLocalizedDescriptionKey: [NSString stringWithFormat:@"Path '%@' does not exist.", path]}];
-        }
-        return nil;
-    }
+    NSNumber *isDirectory;
+    BOOL success = [dirURL getResourceValue:&isDirectory forKey:NSURLIsDirectoryKey error:nil];
+    BOOL isDir = (success && [isDirectory boolValue]);
     
     if (!isDir)
     {
@@ -48,7 +40,7 @@ NSString* const MocktailErrorDomain = @"Mocktail";
     
     // Read the content of the directory
     NSError *bError = nil;
-    NSURL *dirURL = [NSURL fileURLWithPath:dirPath];
+    NSFileManager *fileManager = [NSFileManager defaultManager];
     NSArray *fileURLs = [fileManager contentsOfDirectoryAtURL:dirURL includingPropertiesForKeys:nil options:0 error:&bError];
 
     if (bError)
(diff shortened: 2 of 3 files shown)


OHHTTPStubs/Sources/OHHTTPStubs+Mocktail.m
```diff
@@ -0,0 +1,125 @@
...
+//
+
+#import "OHHTTPStubs+Mocktail.h"
+
+@class OHHTTPStubsDescriptor;
+
+@implementation OHHTTPStubs (Mocktail)
+
+
++(NSArray *)stubRequestsUsingMocktailsAt:(NSURL *)dirURL{
```
AliSoftware: You should use a more idiomatic name for this method, like `stubRequestsUsingMocktailsAtPath:`

OHHTTPStubs/Sources/OHHTTPStubs+Mocktail.m
```diff
@@ -0,0 +1,125 @@
...
+    NSArray *lines = [headerMatter componentsSeparatedByString:@"\n"];
+    if ([lines count] < 4) {
+        if (error) {
+            error = [NSError errorWithDomain:@"Mocktail" code:0 userInfo:@{NSLocalizedDescriptionKey:[NSString stringWithFormat:@"Invalid amount of lines: %u", (unsigned)[lines count]]}];
+        }
+        return nil;
+    }
+    
+    /*handle Mocktail format, adapted from Mocktail implementation, for more details on the file format, check out: https://github.com/square/objc-Mocktail*/
+    NSRegularExpression *methodRegex = [NSRegularExpression regularExpressionWithPattern:lines[0] options:NSRegularExpressionCaseInsensitive error:nil];
```
AliSoftware: You shouldn't ignore the potential `error` here, especially since this `NSRegularExpression` is not hardcoded as a string literal but comes from a file so is highly prone to errors.

OHHTTPStubs/Sources/OHHTTPStubs+Mocktail.m
```diff
@@ -0,0 +1,125 @@
...
+    
+    NSRegularExpression *absoluteURLRegex = [NSRegularExpression regularExpressionWithPattern:lines[1] options:NSRegularExpressionCaseInsensitive error:nil];
```
AliSoftware: ditto

OHHTTPStubs/Sources/OHHTTPStubs+Mocktail.m
```diff
@@ -0,0 +1,125 @@
...
+        return nil;
+    }
+    
+    NSScanner *scanner = [NSScanner scannerWithString:contentsOfFile];
+    NSString *headerMatter = nil;
+    [scanner scanUpToString:@"\n\n" intoString:&headerMatter];
+    NSArray *lines = [headerMatter componentsSeparatedByString:@"\n"];
+    if ([lines count] < 4) {
+        if (error) {
+            error = [NSError errorWithDomain:@"Mocktail" code:0 userInfo:@{NSLocalizedDescriptionKey:[NSString stringWithFormat:@"Invalid amount of lines: %u", (unsigned)[lines count]]}];
```
AliSoftware: You should use a constant string `NSString* const MocktailErrorDomain = @"Mocktail"` everywhere you need to create an Mocktail-related `NSError` (instead of using a string literal like here). You could also expose this constant in the header (`extern NSString* const MocktailErrorDomain`) so that consumers of the lib can compare/check the domain of the error appropriately.

OHHTTPStubs/Sources/OHHTTPStubs.m
```diff
@@ -443,6 +444,8 @@ - (void)streamDataForClient:(id<NSURLProtocolClient>)client
             timingInfo.chunkSizePerSlot = ((stubResponse.dataSize/stubResponse.responseTime) * timingInfo.slotTime);
         }
         
+        timingInfo.bodyOffset = stubResponse.bodyOffset;
```
AliSoftware: I'm not sure I like that hack.

1. `bodyOffset` is not really related to some "timing info" and the pace at which the data is sent periodically, so it's not logical to add it to this struct
2. This modifies the library core just to be able to support Mocktail files

If I add support for Mocktail, given that it's not a feature used by many, I'll probably add it as a opt-in submodule in my podspec, so that people don't get useless extra code if they don't use it, and people needing it simply have to explicitly request `OHHTTPStubs/Mocktail` in their Podfile to get the extra feature. This modification does not allow us to do that given that it modifies the core of OHHTTPStubs.

JinlianWang:
Awesome comments. I was too excited to get this in and obviously let some hack slip in. Have it fixed so that Mocktail extension does not require any change in core library. Also make the error more descriptive.

JinlianWang:
Sorry Oliver. It seems like the project file has some conflict while I am trying to correct the target membership. Shall not take much time to fix it. Thanks!

OHHTTPStubs/Sources/OHHTTPStubs+Mocktail.m
```diff
@@ -0,0 +1,150 @@
...
+            if ([methodRegex numberOfMatchesInString:method options:0 range:NSMakeRange(0, method.length)] > 0) {
+                return YES;
+            }
+        }
+        
+        return NO;
+    } withStubResponse:^OHHTTPStubsResponse*(NSURLRequest *request) {
+        OHHTTPStubsResponse *response = [OHHTTPStubsResponse responseWithFileAtPath:fileURL.path
+                                                                         statusCode:(int)statusCode headers:headers];
+        [response.inputStream setProperty:[NSNumber numberWithUnsignedLongLong:([headerMatter dataUsingEncoding:NSUTF8StringEncoding].length + 2)] forKey:NSStreamFileCurrentOffsetKey];
```
AliSoftware: Please use modern ObjC and the boxing expression syntax `@(x)` instead of `[NSNumber numberWithUnsignedLongLong:x]`.

AliSoftware:
Looks nice so far :+1: 

Regarding the Xcode project conflict, given that I haven't pushed any new conflict that I know of between the start of your branch and your additional commits, that shouldn't be a conflict generated by any change on my part.

Maybe rebase to ensure it's based on master. Otherwise you may revert the xcodeproj and add the files manually again if merging the XML of the pbxproj seems too complicated to you.

----

PS : if you're familiar with CocoaPods, it could be nice to add the necessary changes to the `podspec` as well, so that this addition becomes an optional subspec (move the current `s.source_files` into a "Core" subspec, make it the default, and make a "Mocktail" subspec that depends on Core + add the "+Mocktail.h/m" files to its `ss.source_files`)

JinlianWang:
Somehow there is always conflict even if I back the project file all the way back and add them manually. Could you use the latest version of it and it shall work perfectly? 

As for CocoaPods, my current project is not using it and need some time to refresh my memory. Maybe we treat this as a follow-up or it may take someone more familiar with it minutes to do it? Thanks!

JinlianWang:
Finally got it resolved without any conflicted. It appears that it had conflict because of the recent change on the upstream to the project file on July 28th.

JinlianWang:
@AliSoftware Are you waiting for a CocoaPods change from me before merging it?

AliSoftware:
If you have time yeah why not.
I'm actually more waiting for me to have some time to pull it, run it, test it, merge it, podspec it, tag it, release it. Won't probably have some free time for that until next week.

JinlianWang:
@AliSoftware Just got some cycles today, refresh my memory and got the subspec in. I also added the capability so that it can handle base64 encoded body. It is helpful when it comes to simulate the response for image requests or any other kinds of binary download.

JinlianWang:
@AliSoftware Forgot to mention that I created a test project to test the base64 encoded body part@https://github.com/JinlianWang/TestHTTPStubs. I can make similar effort to the demo project if needed and the desirable change is given.

AliSoftware:
Cool!

Sorry family time this weekend and won't be able to look at it until next week but seems promising!

AliSoftware:
Hi @JinlianWang 

I reviewed your code and did some modifications to improve syntax, readability, add nullability annotations, fix potential crashes and code style, and make it compatible with OS X.

But **there are still some issues especially with the fact that your API is not thought for when `OHHTTPStubs` is used as a (Mac or iOS) framework** (as opposed to a static lib), because in case `OHHTTPStubs` is built as a framework (which is more and more the case with iOS8 btw), the current `NSBundle` will be the framework itself (and not the app, or the Unit Test bundle), and so your methods `stubRequestsUsingMocktailNamed:error:` and `stubRequestsUsingMocktailsAtPath:error:` will both fail, searching the `*.tail` files in the framework itself instead of inside the app bundle.

**I'll let you fix this by changing your API**: typically you'll need to pass the bundle as an argument to your method (and you may avoid using `OHPathHelpers` methods in that case which are more meant to be used by the user, at call site). You can pick some inspiration from my `OHHTTPStubsResponse+HTTPMessage.m` file and its `responseNamed:inBundle:` method for example. I then invite you to test your code by running the Unit Tests **from the "OHHTTPStubs Mac Framework" scheme** so that they run in a framework context.


> **Note**: you can retrieve my modifications on your code by running `git pull https://github.com/AliSoftware/OHHTTPStubs.git mocktail` to pull my new commits into your fork and start working from there.


---

Once you've fixed your API so it can work with frameworks, I'm thinking of two more stuff I'd like to do (but are not related to Mocktail directly, so I could do them myself if you want):

* [ ] Move (or add?) the iOS Unit Tests from the "iOS StaticLib" scheme to the "iOS Framework" scheme, so that these iOS Unit Tests are executed when we build the iOS Framework and not simply the static lib, forcing us to run tests in a framework context in the future to avoid similar cases
* [ ] Further split `OHHTTPStubs` in subspecs. That new `Mocktail` subspec is great and made me think that I should maybe also extract the `HTTPMessage` category in its own subspec as well (and maybe the `JSON` category too, but this category should still be included in the default subspec as it's more common to be useful and used by many, as opposed to HTTPMessage and Mocktail)

JinlianWang:
@AliSoftware Just expose NSBundle as one of the arguments for Framework environments. This is quite a learning process. Thanks!

AliSoftware:
[Version `4.2.0` is out](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/4.2.0) and includes your Mocktail support PR :smiley:

JinlianWang:
Awesome. Thanks @AliSoftware for the great work on such a useful lib. Very thoughtful and powerful interface. We love it here.

JinlianWang:
BTW, I am finalizing up some code to recording traffic in Mocktail format files, so that the can be used for development and testing purpose. Will send your way to get your feedback once done.

Issue #127, opened 2015-10-01
Fix the umbrella header


AliSoftware:
It seems to be broken (see #121)

Probably because:

1. The file named `OHHTTPStubs.h`, which should be my umbrella header, is also the header for one of the classes in the framework (actually the main one, but not the only one), so that's not really a header dedicated for umbrella imports
2. When splitting the library into subspecs, that could be optionally opt-in, I have to conditionally import the other headers in that umbrella header, and this logic might not work in all environments / integration contexts

Maybe one possible solution would be to start by having a dedicated umbrella header, even if it's not called `OHHTTPStubs.h` but rather `OHHTTPStubs-Umbrella.h` for example (I think we can override the default umbrella header in the build settings).

jshier:
This may be related to what I'm seeing in Xcode 7.1. In OHHTTPStubs.h, line 32, I get an error: `Include of non-modular header inside framework module OHHTTPStubs.OHHTTPStubs`. This is using CocoaPods as such: 
```ruby
  pod 'OHHTTPStubs/Core'
  pod 'OHHTTPStubs/NSURLSession'
  pod 'OHHTTPStubs/JSON'
  pod 'OHHTTPStubs/OHPathHelpers'
  pod 'OHHTTPStubs/Swift'
```

keith:
We're seeing this error as well.

billyto:
I tried settting "Allow Non-modular includes in Framework Modules" in Build Settings to YES, but no luck, anyone has another workaround?

keith:
I was able to fix this by manually changing the imports in `OHHTTPStubs.h` with:

```
#import "Compatibility.h"
#import "OHHTTPStubsResponse.h"
```

AliSoftware:
@billyto 
> I tried setting "Allow Non-modular includes in Framework Modules" to YES

Yeah, even if that worked, that would be a workaround to cheat the compiler, wouldn't solve the core issue

@keith Ah interesting. I was wondering if it would be necessary to create a custom umbrella header and a custom modulemap to use it, but if this fix of yours is enough I'd say let's do that instead!

keith:
I'm not sure how that affects users not pulling this in via CocoaPods though.

AliSoftware:
[...]

PR #128, opened 2015-10-12
[OHHTTPStubs] Resolved circular dependency when using dynamic framework

aaresu:
Moved import of OHHTTPStubsResponse from .h to .m to resolving circular dependencies when using dynamic framework.

### DIFF
--- a/OHHTTPStubs/Sources/OHHTTPStubs.h
+++ b/OHHTTPStubs/Sources/OHHTTPStubs.h
@@ -29,8 +29,8 @@
 #import <Foundation/Foundation.h>
 
 #import <OHHTTPStubs/Compatibility.h>
-#import <OHHTTPStubs/OHHTTPStubsResponse.h>
 
+@class OHHTTPStubsResponse;
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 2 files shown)


AliSoftware:
Currently `OHHTTPStubs.h` is the umbrella header for the framework, so `OHHTTPStubs.h` is required by Xcode to import all the headers of the framework.

There is an issue about that (#127) and I need to find a way to solve it, probably by specifying a custom umbrella header in the Build Settings instead than the default one (which is the `.h` file having the same name as the framework, hence `OHHTTPStubs.h` here) and configure it right for all use cases:

* When `OHHTTPStubs` is used as a static lib
* When `OHHTTPStubs` is used as a a framework installed manually
* When `OHHTTPStubs` is used as a a framework via Carthage
* When `OHHTTPStubs` is used as a a framework via CocoaPods
* When `OHHTTPStubs` is used as a a framework via CocoaPods and the user only choose to pick some of the pod's subspecs and not all

AliSoftware:
I'm closing this as this fix continue to break for other configurations, esp. frameworks integrated via Carthage vs. CocoaPods vs. Rome vs. manual etc.

I started a branch `fix/umbrella-header` to try and fix it using a more generic approach.

Thx for the effort and the PR though!

PR #129, opened 2015-10-16
Add methods to create `OHHTPStubResponse`s from `NSURL`s

MaxGabriel:
iOS 4 allowed using `NSURL`s instead of `NSString`s for file paths. This PR allows using `NSURL`s to create `OHHTPStubResponse`s, which provides better compatibility with test code already using `NSURL`s.

I copied each test using `responseWithFileAtPath:statusCode:headers:` and used `responseWithURL:statusCode:headers:` there instead. This ends up duplicating a lot of code, but I know some people prefer that for tests, and this seemed in line with the existing tests.

An alternate implementation of this PR could be to implement the `NSURL` methods in terms of `NSString` based file paths, or vice versa.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # OHHTTPStubs — CHANGELOG
 
+## [4.4.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/4.4.0)
+
+* Added two methods for creating `OHHTTPStubsResponse`s from `NSURL`s that represent file system resources ([@MaxGabriel](https://github.com/MaxGabriel), [#129](https://github.com/AliSoftware/OHHTTPStubs/pull/129))
+
 ## [4.3.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/4.3.0)
 
 * Xcode projects updated to Xcode 7.0 Final
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [4.4.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/4.4.0)
 
-* Added two methods for creating `OHHTTPStubsResponse`s from `NSURL`s that represent file system resources ([@MaxGabriel](https://github.com/MaxGabriel), [#129](https://github.com/AliSoftware/OHHTTPStubs/pull/129))
+* Added methods for creating `OHHTTPStubsResponse`s from `NSURL`s that represent file system resources ([@MaxGabriel](https://github.com/MaxGabriel), [#129](https://github.com/AliSoftware/OHHTTPStubs/pull/129))
 
 ## [4.3.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/4.3.0)
 
--- a/OHHTTPStubs/Sources/OHHTTPStubsResponse.h
+++ b/OHHTTPStubs/Sources/OHHTTPStubsResponse.h
@@ -138,6 +138,22 @@ NS_ASSUME_NONNULL_BEGIN
                            statusCode:(int)statusCode
                               headers:(nullable NSDictionary*)httpHeaders;
 
+
+/**
+ *  Builds a response given a URL, the status code, and headers.
+ *
+ *  @param url         The URL for the data to reutrn in the response
+ *  @param statusCode  The HTTP Status Code to use in the response
+ *  @param httpHeaders The HTTP Headers to return in the response
+ *
+ *  @return An `OHHTTPStubsResponse` describing the corresponding response to return by the stub
+ *
+ *  @note This method applies only to URLs that represent file system resources
+ */
++(instancetype)responseWithURL:(NSURL *)url
+                    statusCode:(int)statusCode
+                       headers:(nullable NSDictionary *)httpHeaders;
+
 /* -------------------------------------------------------------------------- */
 #pragma mark > Building an error response
 
@@ -238,6 +254,21 @@ NS_ASSUME_NONNULL_BEGIN
                        statusCode:(int)statusCode
                           headers:(nullable NSDictionary*)httpHeaders;
 
+
+/**
+ *  Initialize a response with a given URL, statusCode and headers.
+ *
+ *  @param url         The URL for the data to reutrn in the response
+ *  @param statusCode  The HTTP Status Code to use in the response
+ *  @param httpHeaders The HTTP Headers to return in the response
+ *
+ *  @return An `OHHTTPStubsResponse` describing the corresponding response to return by the stub
+ *
+ *  @note This method applies only to URLs that represent file system resources
+ */
+-(instancetype)initWithURL:(NSURL *)url
+                statusCode:(int)statusCode
+                   headers:(nullable NSDictionary *)httpHeaders;
 
 /**
  *  Initialize a response with the given data, statusCode and headers.
--- a/OHHTTPStubs/Sources/OHHTTPStubsResponse.h
+++ b/OHHTTPStubs/Sources/OHHTTPStubsResponse.h
@@ -142,7 +142,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Builds a response given a URL, the status code, and headers.
  *
- *  @param url         The URL for the data to reutrn in the response
+ *  @param fileURL     The URL for the data to return in the response
  *  @param statusCode  The HTTP Status Code to use in the response
  *  @param httpHeaders The HTTP Headers to return in the response
  *
@@ -150,9 +150,9 @@ NS_ASSUME_NONNULL_BEGIN
  *
  *  @note This method applies only to URLs that represent file system resources
  */
-+(instancetype)responseWithURL:(NSURL *)url
-                    statusCode:(int)statusCode
-                       headers:(nullable NSDictionary *)httpHeaders;
++(instancetype)responseWithFileURL:(NSURL *)fileURL
+                        statusCode:(int)statusCode
+                           headers:(nullable NSDictionary *)httpHeaders;
 
 /* -------------------------------------------------------------------------- */
 #pragma mark > Building an error response
@@ -258,7 +258,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Initialize a response with a given URL, statusCode and headers.
  *
- *  @param url         The URL for the data to reutrn in the response
+ *  @param fileURL     The URL for the data to return in the response
  *  @param statusCode  The HTTP Status Code to use in the response
  *  @param httpHeaders The HTTP Headers to return in the response
  *
@@ -266,9 +266,9 @@ NS_ASSUME_NONNULL_BEGIN
  *
  *  @note This method applies only to URLs that represent file system resources
  */
--(instancetype)initWithURL:(NSURL *)url
-                statusCode:(int)statusCode
-                   headers:(nullable NSDictionary *)httpHeaders;
+-(instancetype)initWithFileURL:(NSURL *)fileURL
+                    statusCode:(int)statusCode
+                       headers:(nullable NSDictionary *)httpHeaders;
 
 /**
  *  Initialize a response with the given data, statusCode and headers.
--- a/OHHTTPStubs/Sources/OHHTTPStubsResponse.m
+++ b/OHHTTPStubs/Sources/OHHTTPStubsResponse.m
@@ -157,7 +157,7 @@ const double OHHTTPStubsDownloadSpeedWifi   =- 12000 / 8; // kbps -> KB/s
     NSInputStream* inputStream;
     if (fileURL)
     {
-        NSAssert(fileURL.fileURL, @"Only file URLs may be passed to %@", NSStringFromSelector(_cmd));
+        NSAssert(fileURL.isFileURL, @"Only file URLs may be passed to %@", NSStringFromSelector(_cmd));
         inputStream = [NSInputStream inputStreamWithURL:fileURL];
     }
     else
--- a/OHHTTPStubs/Sources/OHHTTPStubsResponse.m
+++ b/OHHTTPStubs/Sources/OHHTTPStubsResponse.m
@@ -75,6 +75,15 @@ const double OHHTTPStubsDownloadSpeedWifi   =- 12000 / 8; // kbps -> KB/s
     return response;
 }
 
++(instancetype)responseWithURL:(NSURL *)url
+                    statusCode:(int)statusCode
+                       headers:(nullable NSDictionary *)httpHeaders
+{
+    OHHTTPStubsResponse* response = [[self alloc] initWithURL:url
+                                                   statusCode:statusCode
+                                                      headers:httpHeaders];
+    return response;
+}
 
 #pragma mark > Building an error response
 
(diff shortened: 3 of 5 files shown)


OHHTTPStubs/Sources/OHHTTPStubsResponse.h
```diff
@@ -138,6 +138,22 @@ NS_ASSUME_NONNULL_BEGIN
                            statusCode:(int)statusCode
                               headers:(nullable NSDictionary*)httpHeaders;
 
+
+/**
+ *  Builds a response given a URL, the status code, and headers.
+ *
+ *  @param url         The URL for the data to reutrn in the response
```
AliSoftware: `reutrn`

AliSoftware:
Looks good to me :+1: Thx for the PR.

I'd just suggest to maybe rename the methods `withFileURL:` instead of just `withURL:` so that people don't think they can use `NSURL`s pointing to network resources in there (which could get confusing especially in the context of `OHHTTPStubs` and network requests :smile: ) but only using fileURLs.

AliSoftware:
Also, once you've made the suggested changes, could you add an entry in the `CHANGELOG` to credit yourself? Thx!

MaxGabriel:
@AliSoftware I've made the suggested changes. I can rebase them into a single commit if you'd like; just wanted to give you the chance to see the new commit first.

OHHTTPStubs/UnitTests/Test Suites/NilValuesTests.m
```diff
@@ -142,6 +170,32 @@ - (void)test_InvalidPath
     [self waitForExpectationsWithTimeout:kResponseTimeTolerence handler:nil];
 }
 
+- (void)test_InvalidPathWithURL
+{
+    [OHHTTPStubs stubRequestsPassingTest:^BOOL(NSURLRequest *request) {
+        return YES;
+    } withStubResponse:^OHHTTPStubsResponse *(NSURLRequest *request) {
+        return [[OHHTTPStubsResponse responseWithFileURL:[NSURL URLWithString:@"-invalid-url"] statusCode:501 headers:nil]
```
AliSoftware: Mmmh actually just realized that `[NSURL URLWithString:@"-invalid-url"]` will return `nil` and thus this test is kind of a duplicate of `test_NilPathWithURL`.

Probably better use a valid URL (as of the RFC, so it doesn't return `nil`) but that resolves to an invalid resource (non-existing file) instead?

AliSoftware:
Great work, thanks! :+1:

I don't mind keeping multiple commits, no need to rebase into a single one. I usually like to create a _merge commit_ anyway, so it'll be clear in the git graph that those commits go together.

Just some last requests, it would be useful to:
* transform your `test_InvalidPathWithURL` test so it uses a valid URL but pointing to an invalid path (`file://foo/bar` for example) instead of it using something that will use a `nil` URL in the end
* add a test for an `NSURL` for which you couldn't get the file size, to cover the case of [this part of your code](https://github.com/AliSoftware/OHHTTPStubs/pull/129/files#diff-f9aee06c9dbb0fc6a44c5b9bccb7a14fR186)

Sorry for the extra work and for not pointing to that previously :relaxed:

AliSoftware:
> An alternate implementation of this PR could be to implement the NSURL methods in terms of NSString based file paths, or vice versa.

Ah, good point. Maybe we should then change the implementation of the methods using NSString so they now create an  `[NSURL URLWithFileAtPath:]` and call you implementation instead, to avoid code duplication in those implementation?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # OHHTTPStubs — CHANGELOG
 
+## [4.4.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/4.4.0)
+
+* Added two methods for creating `OHHTTPStubsResponse`s from `NSURL`s that represent file system resources ([@MaxGabriel](https://github.com/MaxGabriel), [#129](https://github.com/AliSoftware/OHHTTPStubs/pull/129))
```
AliSoftware: Probably just say `added methods for…` (you actually added more than two :wink:)

AliSoftware:
Additional idea, not sure it's really useful/needed, but do you think adding an [`OHPathHelper`](https://github.com/AliSoftware/OHHTTPStubs/blob/master/OHHTTPStubs/Sources/OHPathHelpers/OHPathHelpers.m) for returning `NSURL` for fixtures would also be handy for people using `NSURL`s in their tests?

MaxGabriel:
> Additional idea, not sure it's really useful/needed, but do you think adding an OHPathHelper for returning NSURL for fixtures would also be handy for people using NSURLs in their tests?

I'm not sure. Right now I have two methods that all my test code uses to load files for HTTP stubbing (pasted below), so it's not a considerable typing saver for me. They'd be quick to write so I'll add them if you'd like.

```objc
+ (NSMutableDictionary *)jsonForResource:(NSString *)resource
{
    NSURL *url = [[NSBundle bundleForClass:[self class]] URLForResource:resource withExtension:@"json"];
    NSData *data = [NSData dataWithContentsOfURL:url];
    
    NSError *error;
    NSMutableDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:NSJSONReadingMutableContainers|NSJSONReadingMutableLeaves error:&error];
    
    NSAssert(error == nil, @"Error should be nil");
    NSAssert(json, @"JSON shouldn't be nil");
    
    return json;
}
```

```objc
+ (id<OHHTTPStubsDescriptor>)stubRequestContainingString:(NSString *)string withVideoFile:(NSString *)videoFileName {
    NSParameterAssert(string);
    NSParameterAssert(videoFileName);
    NSURL *url = [[NSBundle bundleForClass:[self class]] URLForResource:videoFileName withExtension:@"mp4"];
    NSParameterAssert(url);
    
    return [OHHTTPStubs stubRequestsPassingTest:^BOOL(NSURLRequest *request) {
        return [request.URL.absoluteString hzContainsString:string];
    } withStubResponse:^OHHTTPStubsResponse *(NSURLRequest *request) {
        return [[OHHTTPStubsResponse alloc] initWithURL:url statusCode:200 headers:@{@"content-type":@"video/mp4"}];
    }];
}
```

AliSoftware:
Yeah I'm not sure they'd be needed, anyone can add them in their code anyway, that'd just be a on-liner, so let's not bother about that.

MaxGabriel:
Awesome, thanks @AliSoftware. I think all comments are responded to. I wanted to note that these two points ended up both being covered by `test_InvalidPathWithURL`.

> * transform your test_InvalidPathWithURL test so it uses a valid URL but pointing to an invalid path (file://foo/bar for example) — instead of it using a string that will return a nil NSURL and duplicate the previous test
> * add a test for an NSURL for which you couldn't get the file size, to cover the case of this part of your code (maybe using a network NSURL there?)

AliSoftware:
Awesome work man! :tada: 
Could you manage to add one last test that ensure you fail properly (we hit your `NSAssert`) if we give a non-file NSURL?

I'm going to bed now, and will probably merge it tomorrow. (Will probably need to fix #127 first before doing a new official release though)

MaxGabriel:
Sure, good idea for that test.

MaxGabriel:
@AliSoftware Any idea why these [tests fail](https://travis-ci.org/AliSoftware/OHHTTPStubs/jobs/85849646)? It looks like `[NSURL URLWithString:@"http://www.iana.org/domains/example/"];` is being considered as a file URL by the tests. The tests pass for me locally, but not on Travis.

MaxGabriel:
Potentially related is that `isFileURL` is only available on iOS 8+, whereas `OHHTPStubs` supports back to iOS 5. Thoughts?

OHHTTPStubs/Sources/OHHTTPStubsResponse.m
```diff
@@ -135,23 +144,47 @@ -(instancetype)initWithFileAtPath:(NSString*)filePath
...
+    if (error) {
+        NSLog(@"%s: Error getting file size for URL. Returning empty data", __PRETTY_FUNCTION__);
+        inputStream = [NSInputStream inputStreamWithData:[NSData data]];
+    } else {
+        self = [self initWithInputStream:inputStream
+                                dataSize:[fileSize unsignedLongLongValue]
+                              statusCode:statusCode
+                                 headers:httpHeaders];
+    }
+
```
samirGuerdah: I think that you should check the return value of the method getResourceValue and not the error and check also the availability of the property NSURLFileSizeKey if the method returns success. 

From the Apple Documentation:
"If this method returns YES and the value is populated with nil, it means that the resource property is not available for the specified resource, and that no errors occurred when determining that the resource property was unavailable."
``` objective-c
BOOL success = [fileURL getResource...]
if (success && fileSize) { }
else {} // Populate with [NSData data]
```

AliSoftware: Good point @samirGuerdah !

MaxGabriel: Oh wow, actually this method is totally incorrect; in the case of an error the object is never initialized. 

Do you think something like this would work? I think I only need to check that the `NSNumber` is `nil`.

```objc
NSError *error;
NSNumber *fileSize;
BOOL success = [fileURL getResourceValue:&fileSize forKey:NSURLFileSizeKey error:&error];

if (!fileSize) {
    fileSize = @0;
    inputStream = [NSInputStream inputStreamWithData:[NSData data]];
    NSLog(@"%s: Couldn't get file size for URL; returning empty data instead. \
The URL was: %@. \
The operation to retrieve the file size was %@. \
The error associated with that operation was: %@",
    __PRETTY_FUNCTION__, fileURL, success ? @"successful" : @"unsuccessful", error);
}
    
self = [self initWithInputStream:inputStream
                        dataSize:[fileSize unsignedLongLongValue]
                      statusCode:statusCode
                         headers:httpHeaders];
```

AliSoftware: You don't even seem to check whether `success` is `YES` in this last implementation?

MaxGabriel: @AliSoftware From the documentation it sounds like `success` can be `YES`, no error would be set, and the returned value would be nil:

> If this method returns YES and value is populated with nil, it means the resource property is not available for the specified resource and no errors occurred when determining the resource property was not available.

This code really only cares about `fileSize` being non-nil—I think the values of `success` and `error` are only useful for informing the user why the operation failed in this case.

AliSoftware: Well, I'm not so sure. I think for some bizarre reasons the method could also fail while setting a non-nil value to `fileSize`. I mean, sure that's very unlikely of course, and would be strange. But nothing from the doc prevent it from happening.

So if the method return a success boolean, you're always supposed to test it anyway. That's the official pattern for all those methods taking an `NSError**` pioneer and returning a `BOOL`, you're supposed to test the returned `BOOL`, if it's `NO` then use the `NSError` to display the message, and if it's `YES` proceed with the data. In those methods, you're not supposed to test if `NSError` is `nil` or not to see if there is an error, but to test the returned `BOOL`. So I'd say that's the same for `fileSize`.

That's probably just style indeed because I don't think the method would even fill the `fileSize` pointer but still return `NO`, but that's how the method is defined and how the methods using `NSError**` and byref parameters are supposed to be used.

MaxGabriel: Fair enough. Here's what I have then:

```objc
-(instancetype)initWithFileURL:(NSURL *)fileURL
                    statusCode:(int)statusCode
                       headers:(nullable NSDictionary *)httpHeaders {
    NSInputStream* inputStream;
    NSNumber *fileSize;
    if (fileURL)
    {
        // [NSURL -isFileURL] is only available on iOS 8+
        if (![fileURL.scheme isEqualToString:NSURLFileScheme]) {
            @throw [NSException exceptionWithName:NSInternalInconsistencyException
                                           reason:[NSString stringWithFormat:@"Only file URLs may be passed to %@", NSStringFromSelector(_cmd)]
                                         userInfo:nil];
        }
        inputStream = [NSInputStream inputStreamWithURL:fileURL];

        NSError *error;
        BOOL success = [fileURL getResourceValue:&fileSize forKey:NSURLFileSizeKey error:&error];
        
        if (!success || !fileSize) {
            fileSize = @0;
            inputStream = [NSInputStream inputStreamWithData:[NSData data]];
            NSLog(@"%s: Couldn't get file size for URL; returning empty data instead. \
The URL was: %@. \
The operation to retrieve the file size was %@. \
The error associated with that operation was: %@",
                  __PRETTY_FUNCTION__, fileURL, success ? @"successful" : @"unsuccessful", error);
        }
    }
    else
    {
        NSLog(@"%s: nil URL. Returning empty data", __PRETTY_FUNCTION__);
        fileSize = @0;
        inputStream = [NSInputStream inputStreamWithData:[NSData data]];
    }
    
    
    
    self = [self initWithInputStream:inputStream
                            dataSize:[fileSize unsignedLongLongValue]
                          statusCode:statusCode
                             headers:httpHeaders];
    
    return self;
}
```

AliSoftware: Feels odd to me that we'd `@throw` in one case (using a non-file URL) and just `NSLog` in the other (failing to get file size). I say we'd use one or the other for both. Probably `@throw`. What's your take?

MaxGabriel: I like this version a better; there's some duplication in returning empty data but there's much less nesting:

```objc
-(instancetype)initWithFileURL:(NSURL *)fileURL
                    statusCode:(int)statusCode
                       headers:(nullable NSDictionary *)httpHeaders {
    if (!fileURL) {
        NSLog(@"%s: nil URL. Returning empty data", __PRETTY_FUNCTION__);
        return [self initWithInputStream:[NSInputStream inputStreamWithData:[NSData data]]
                         dataSize:0
                       statusCode:statusCode
                          headers:httpHeaders];
    }
    
    // [NSURL -isFileURL] is only available on iOS 8+
    if (![fileURL.scheme isEqualToString:NSURLFileScheme]) {
        @throw [NSException exceptionWithName:NSInternalInconsistencyException
                                       reason:[NSString stringWithFormat:@"Only file URLs may be passed to %@",NSStringFromSelector(_cmd)]
                                     userInfo:nil];
    }
    
    NSNumber *fileSize;
    NSError *error;
    const BOOL success = [fileURL getResourceValue:&fileSize forKey:NSURLFileSizeKey error:&error];
    
    if (!success || !fileSize) {
        NSLog(@"%s: Couldn't get file size for URL; returning empty data instead. \
The URL was: %@. \
The operation to retrieve the file size was %@. \
The error associated with that operation was: %@",
              __PRETTY_FUNCTION__, fileURL, success ? @"successful" : @"unsuccessful", error);
        return [self initWithInputStream:[NSInputStream inputStreamWithData:[NSData data]]
                                dataSize:0
                              statusCode:statusCode
                                 headers:httpHeaders];
    }
    
    return [self initWithInputStream:[NSInputStream inputStreamWithURL:fileURL]
                            dataSize:[fileSize unsignedLongLongValue]
                          statusCode:statusCode
                             headers:httpHeaders];
}
```

MaxGabriel: > Feels odd to me that we'd @throw in one case (using a non-file URL) and just NSLog in the other (failing to get file size). I say we'd use one or the other for both. Probably @throw. What's your take?

@AliSoftware I definitely agree it's strange; I prefer `@throw` for both, especially since this is a testing library. The only thing I was worried about was that existing code could be counting on `nil` file paths returning empty data instead of throwing an exception.

AliSoftware: I think a `nil` URL should still return empty `[NSData data]`, because it makes sense and because it's already the behavior for what we have for the path.
But any misuse of the method like using a malformed URL, a non-file URL, or an URL for which we can't determine the resource size should `@throw`.

AliSoftware:
Seems like tests are failing now :-/

MaxGabriel:
@AliSoftware Oh good call about the assertions. It looks like assertions are disabled in the release configuration of the libraries. Do you prefer using `@throw` or just setting `ENABLE_NS_ASSERTIONS` to `YES`? There are a couple of other `NSAssert`s in the codebase besides the ones I added.

AliSoftware:
Good question. I think keeping `NSAssert` makes more sense, as it's what this is for after all. But I guess this means that we need to make the Unit Tests test the Debug build — where assertions are enabled — not the Release one — where they are blocked which is probably a good thing too.

AliSoftware:
Yay! Champagne! :tada: 
I thing that's all good now, let's merge this at last!

Thanks so much for the great job and time spent on this @MaxGabriel :beers:

MaxGabriel:
Thanks for all your help @AliSoftware and @samirGuerdah! I think this came out great

PR #134, opened 2015-10-28
Experimental/tv os support

tiagomartinho:
Hello! 

I created the branch to support tvOS.

It uses the pod AFNetworking 3.0.0 (beta 1) since they added support to tvOS.

I only added the target for building a Framework and didn't add the target for a static library.

I had to bump the deployment target of iOS to 7.0 and OS X to 10.9 because of AFNetworking 3.

Two tests don't pass (and I don't understand their usefulness) since the method setRedirectResponseBlock is no longer defined in AFNetworking.


AliSoftware:
Thanks for the PR !  :apple:  :tv:

Sadly I won't have much time those days to fully review it but I hope to look at it in detail this week-end.

I'm not sure I understood why you had to bump the minimum deployment targets of the pod for iOS and OS X, since the pod doesn't have a dependency on AFNetworking, only the test targets do. 
So theoretically you should only had to bump the Deployment Target build setting of the Unit Tests target, not the podspec which doesn't depend on AFN, right?

Regarding the tests about the redirect block was to ensure the redirections were properly followed by `OHHTTPStubs` and especially that there weren't any regression on [this part of the code logic](https://github.com/AliSoftware/OHHTTPStubs/blob/master/OHHTTPStubs/Sources/OHHTTPStubs.m#L347-L366). So I think we should still keep those tests, but they probably need to be adapted/rewritten to make a redirection using whatever new method AFNetworking 3 provides for that.

tiagomartinho:
You were right, the deployment version should only be increased by the tests targets. I reverted the deployment target for the pod targets.

Regarding the tests I found that AFNetworking has the method setTaskWillPerformHTTPRedirectionBlock in AFURLSessionManager. I will try using this one to adapt the failing tests.

tiagomartinho:
I was able to fix the failing tests using the setTaskWillPerformHTTPRedirectionBlock method from AFNetworking.

AliSoftware:
Great! Will take a look probably this weekend. Thanks a lot!

AliSoftware:
[4.5.0 is out](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/4.5.0), thanks a lot for this PR again!

tiagomartinho:
Welcome! Thank you for merging.

PR #140, opened 2015-11-12
Use explicit CFBundleVersion number

siemensikkema:
When building OHHTTPStubs using Carthage the version number ended up being empty. This caused an issue when uploading my app to iTunes Connect:

`ERROR ITMS-90056: This bundle is invalid. The Info.plist file is missing the required key: CFBundleVersion`

Hard-coding the version resolved the issue for me. The file `OHHTTPStubs Mac-Info.plist` already has the CFBundleVersion set to '1'.

### DIFF
--- a/OHHTTPStubs/Supporting Files/OHHTTPStubs iOS-Info.plist
+++ b/OHHTTPStubs/Supporting Files/OHHTTPStubs iOS-Info.plist
@@ -19,7 +19,7 @@
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<string>1</string>
 	<key>NSPrincipalClass</key>
 	<string></string>
 </dict>


AliSoftware:
Thanks for the PR !

The problem with this fix is that the framework will always be flagged as version "1" instead of the real version. When `OHHTTPStubs` increment its version (in the podspec and in the frameworks) the new version will not change the version in the framework. For people relying on the version of the framework on their code, and for proper dependency and version management, this is not good.

The proper fix is probably properly define the `CURRENT_PROJECT_VERSION` build setting (which is normally done by Xcode automatically so something probably went wrong elsewhere)

siemensikkema:
Sounds plausible. I don't know why this wasn't set correctly. I have other dependencies where this doesn't cause a problem. Could it be related to the renaming of `OHHTTPStubs iOS-Info.plist` to `Info.plist`? I don't see this in my other dependencies but this is just a wild guess.
Some use `${CURRENT_PROJECT_VERSION}` instead of `$(CURRENT_PROJECT_VERSION)` but either seems to work.
You might want to set the version for Mac to `$(CURRENT_PROJECT_VERSION)` as well because currently it reads `1`.

AliSoftware:
Fixed by commit 69c64dc6cf9d5df9914aaed027c6a37a49788d39

Thanks for pointing that out and sorry it took me so long to fix that.
Please test version 4.7.1 and ensure it indeed fixed your issue with versioning & iTunesConnect :wink:

siemensikkema:
Great! I can stop using my fork now :) I'll upload a test build soon and if I run into any problems I'll report back.

PR #144, opened 2015-12-15
add `nullable` specifier to block that can be nil

DerLobi:
The documentation comment for `-onStubActivation:` says that the block can be `nil`:
> Set it to `nil` to do nothing. Defaults is `nil`

The file is within `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` so I get a warning 
> :warning: Null passed to a callee that requires a non-null argument

when I pass `nil` for the block parameter.
I have added `nullable` to the method signature to avoid the warning.


AliSoftware:
Nice catch!

Maybe we should also add `nullable` to the attributes of the `@property(…)` [here in the class extension](https://github.com/AliSoftware/OHHTTPStubs/blob/master/OHHTTPStubs/Sources/OHHTTPStubs.m#L47) too? (even if the most important part is in the public API, but just to be consistent)

Otherwise :+1:

DerLobi:
Added it to the property

AliSoftware:
:ok_hand: Thanks again for the PR!

Issue #139, opened 2015-11-11
Unable to access whether stubs are currently enabled or not


engineeringman:
I'd like to add a runtime window that will enable us to toggle the enabled state of OHHTTPStubs at runtime.  However there's no way to know whether OHHTTPStubs is currently enabled or not, so I have no idea of which state to set my toggle switch on startup without making assumptions.  This seems like a property that should exist.

AliSoftware:
Sorry not to have responded earlier, I got tons of stuff to do and forgot this.

That could definitely be a property to add. Won't have much time to add it soon though, but don't hesitate to make a PR to add it, that shouldn't be too complicated to implement :wink:

jzucker2:
I saw this and decided to take a run at the idea. Let me know what you think https://github.com/AliSoftware/OHHTTPStubs/pull/159

jzucker2:
Just in case you forgot, you can close this now :)

AliSoftware:
Good poin, thanks!

PR #159, opened 2016-03-08
Added isEnabled getters for OHHTTPStubs.

jzucker2:
Reflects current implementation. This branch does not update the version or run `pod update` for the examples. There is a minor bug in Cocoapods for building projects for different platforms. Not sure how you build your projects so I can add that to the branch when I know what to do. This is in response to https://github.com/AliSoftware/OHHTTPStubs/issues/139

### DIFF
--- a/OHHTTPStubs/Sources/OHHTTPStubs.m
+++ b/OHHTTPStubs/Sources/OHHTTPStubs.m
@@ -44,6 +44,7 @@ static NSTimeInterval const kSlotTime = 0.25; // Must be >0. We will send a chun
 @interface OHHTTPStubs()
 + (instancetype)sharedInstance;
 @property(atomic, copy) NSMutableArray* stubDescriptors;
+@property(atomic, copy) NSNumber* enabledStateNumber;
 @property(atomic, copy, nullable) void (^onStubActivationBlock)(NSURLRequest*, id<OHHTTPStubsDescriptor>);
 @end
 
@@ -105,7 +106,7 @@ static NSTimeInterval const kSlotTime = 0.25; // Must be >0. We will send a chun
 {
     if (self == [OHHTTPStubs class])
     {
-        [self setEnabled:YES];
+        [self _setEnable:YES];
     }
 }
 - (instancetype)init
@@ -114,13 +115,14 @@ static NSTimeInterval const kSlotTime = 0.25; // Must be >0. We will send a chun
     if (self)
     {
         _stubDescriptors = [NSMutableArray array];
+        _enabledStateNumber = @YES; // assume initialize has already been run
     }
     return self;
 }
 
 - (void)dealloc
 {
-    [self.class setEnabled:NO];
+    [self.class _setEnable:NO];
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


OHHTTPStubs/Sources/OHHTTPStubs.m
```diff
@@ -149,9 +149,11 @@ +(void)removeAllStubs
 
 #pragma mark > Disabling & Re-Enabling stubs
 
+static BOOL currentEnabledState = NO;
```
AliSoftware: global variables are not thread safe, so this is not an acceptable solution.

AliSoftware:
What's the CP issue/error message you're encountering?

jzucker2:
This is a thread about the issue. Supposedly fixed in the next beta. Definitely present in 0.39 https://github.com/CocoaPods/CocoaPods/issues/4478

AliSoftware:
Oh I see. Never had that issue before, it passed when running OHHTTPStubs tests on CI before which runs the checks for each platform, so will take a look later :+1:

(PS: a bit busy this week so I might not be able to test this PR until this weekend)

OHHTTPStubs/Sources/OHHTTPStubs.m
jzucker2: Thanks for the suggestion!

jzucker2:
No rush on this pull request. Just saw an opportunity to help and wanted to contribute.

AliSoftware:
:+1:

jzucker2:
There's not a ton of great documentation on @synchronized, seems it's slightly out of favor. I tried to follow your pattern in this implementation. Also open to using a NSLock or a mutex, if you'd prefer. Just figured I'd try out your original suggestion first.

OHHTTPStubs/Sources/OHHTTPStubs.m
```diff
@@ -208,6 +243,25 @@ +(void)onStubActivation:( nullable void(^)(NSURLRequest* request, id<OHHTTPStubs
 ////////////////////////////////////////////////////////////////////////////////
 #pragma mark - Private instance methods
 
+-(BOOL)isEnabled
+{
+    BOOL enabled = NO;
+    @synchronized(self)
```
jzucker2: Using `self` instead of `_enabledStateNumber` because that might change and cause multiple locks to be enabled for this property

OHHTTPStubs/Sources/OHHTTPStubs.m
```diff
@@ -208,6 +243,25 @@ +(void)onStubActivation:( nullable void(^)(NSURLRequest* request, id<OHHTTPStubs
...
+    {
+        enabled = _enabledStateNumber.boolValue;
+    }
+    return enabled;
+}
+
+-(void)setEnabled:(BOOL)enable
+{
+    @synchronized(self)
```
jzucker2: Same as above

AliSoftware:
Yeah, OHHTTPStubs starts having history, that code was written a while back when `@synchronized` was the way to go. Nowadays if I were to re-do it, I'd probably use `NSLock` or `OSSpinLock` or something. But as this is not the purpose of this PR, I think it's better to keep `@synhronized` here for now. We can still decide to make the switch for `NSLock` or `OSSpinLock` in a separate and dedicated PR if we feel the need, but that's a separate issue :wink:

OHHTTPStubs/Sources/OHHTTPStubs.m
```diff
@@ -44,6 +44,7 @@ @interface OHHTTPStubsProtocol : NSURLProtocol @end
 @interface OHHTTPStubs()
 + (instancetype)sharedInstance;
 @property(atomic, copy) NSMutableArray* stubDescriptors;
+@property(atomic, copy) NSNumber* enabledStateNumber;
```
AliSoftware: Why not using a `BOOL`, as it will be initialized to `YES` at init time so you don't really need the `YES/NO/nil` tri-state anyway?

OHHTTPStubs/Sources/OHHTTPStubs.m
```diff
@@ -189,6 +199,31 @@ + (void)setEnabled:(BOOL)enable forSessionConfiguration:(NSURLSessionConfigurati
...
+
++ (BOOL)isEnabledForSessionConfiguration:(NSURLSessionConfiguration *)sessionConfig
+{
+    // Runtime check to make sure the API is available on this version
+    if (   [sessionConfig respondsToSelector:@selector(protocolClasses)]
+        && [sessionConfig respondsToSelector:@selector(setProtocolClasses:)])
+    {
+        NSMutableArray * urlProtocolClasses = [NSMutableArray arrayWithArray:sessionConfig.protocolClasses];
+        Class protoCls = OHHTTPStubsProtocol.class;
+        if ([urlProtocolClasses containsObject:protoCls])
```
AliSoftware: Why not directly `return [urlProtocolClasses containsObject:protoCls];`?

jzucker2: Good point!

jzucker2:
Fixed!

AliSoftware:
Awesome, thanks again for the PR :ok_hand:

PR #161, opened 2016-03-18
Adds more logging blocks for debugging and extended features

jzucker2:
This isn't a large change, but I was hoping for a little more insight into when certain actions occur inside OHHTTPStubs so I added some more blocks in addition to `onStubActivationBlock`. Now there is also a `onStubRedirectBlock` and a `afterStubFinishBlock` for better clarity. As well, I pass in the response used by the stub. In my case I use single stub that generates multiple different responses, similar to your example [Different responses for each call](https://github.com/AliSoftware/OHHTTPStubs/wiki/Usage-Examples#different-responses-for-each-call) and so it was easier to track the response returned than the descriptor that generates that response.

I'm totally open to a discussion on this, but it'd be really helpful for me if something like this was added, hence why I put together this PR.


jzucker2:
Also, was CI disabled for this project?

AliSoftware:
Thanks a lot for this PR! Looks quite good! :+1: 

I don't know why Travis didn't pick up that PR, it usually does ([it did for the last PR 160, so why not this one?](https://travis-ci.org/AliSoftware/OHHTTPStubs/pull_requests)), so that's strange indeed.

Could you also add an entry in the `CHANGELOG.md` to list your changes there and credit yourself? Thanks!

jzucker2:
CI ran this time!

jzucker2:
Thank you! That's an easy fix

Issue #52, opened 2014-02-06
Body of request never makes it to stubRequestPassingTest:


jpalten:
I use OHHTTPStubs to test if my AFHTTPRequest send the correct xml as body in a request. This works fine and all tests pass, checking the body that would be sent in the stubRequestPassingTest:

Now I moved to using AFHTTPSessionManager, and the program still works fine, but the tests fail, telling me that the body to be sent is empty, instead of the expected xml.

Something funny that happens: the request that gets prepared by my program has a different pointer (0xe1efbc0) than the request (0xe2c3310) that is checked in the stub. Any idea what is going on? Is it possible the requests gets copied, but the body didn't make it to the copy?

Here is a log from my console:

2014-02-06 14:09:36.278 xctest[69495:303] Posting request 0xe1efbc0 with body: <ajax-request action="getstat" comp="system"></ajax-request>
2014-02-06 14:09:36.278 xctest[69495:1903] Stubbing url /admin/_cmdstat.jsp
2014-02-06 14:09:36.278 xctest[69495:1903] checking body of request 0xe2c3310
2014-02-06 14:09:36.279 xctest[69495:1903] Would sent body: 
/Volumes/InternalHD/Projecten/ZoneDirectorViewer/ZoneDirectorViewer/Connection/ServerConnectionTests.m:551: error: -[ServerConnectionTests testGetSystemInfo] : ((requestedBody) equal to (expectedBody)) failed: ("") is not equal to ("<ajax-request action="getstat" comp="system"></ajax-request>")

samskiter:
Hi, not using OHHTTPStubs, but stumbled into this while hitting exactly this issue myself. Found your discussion very useful for diagnosing the problem. Boiled it down and produced a project demonstrating the problem at it's simplest (and the difference between NSURLSession and NSURLConnection) - https://github.com/samskiter/testnsurlprotocol

Thanks for the information everyone and thanks @AliSoftware for the workaround

AliSoftware:
@samskiter thx for the feedback :+1:

AliSoftware:
Hi all, this issue should now have a workaround thanks to #166 (which has just been merged into master)!
You'll now be able to use the new `OHHTTPStubs_HTTPBody` property on `NSURLRequest` if you want to test your request body in stubs even after iOS cleared the native property now.

Will update the wiki entry and do a release soon.

PR #166, opened 2016-03-31
Adds support for HTTPBody during testing when using NSURLSession

felixLam:
This addresses the issue described in https://github.com/AliSoftware/OHHTTPStubs/wiki/Testing-for-the-request-body-in-your-stubs and #52 

It also avoids unnecessary changes to the code under test as the setter and getter for `HTTPBody` can still be used without change in behavior (as far as one can tell). 

The new getter `-[NSURLRequest OHHTTPStubs_HTTPBody]` can be used in your `[OHHTTPStubs stubRequestsPassingTest:withStubResponse:]` to test the `HTTPBody`.

One known issue is that if you set the HTTPBody and later reset it to nil (for whatever reason),`OHHTTPStubs_HTTPBody` will still return the last non-nil value.   

Thanks to @shagedorn for the help on method swizzling.

### DIFF
--- a/OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.h
+++ b/OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.h
@@ -1,0 +1,48 @@
+/***********************************************************************************
+*
+* Copyright (c) 2016 Sebastian Hagedorn, Felix Lamouroux
+*
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
+* in the Software without restriction, including without limitation the rights
+* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the Software is
+* furnished to do so, subject to the following conditions:
+*
+* The above copyright notice and this permission notice shall be included in
+* all copies or substantial portions of the Software.
+*
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+* THE SOFTWARE.
+*
+***********************************************************************************/
+
+////////////////////////////////////////////////////////////////////////////////
+#pragma mark - Imports
+
+#import <Foundation/Foundation.h>
+
+// This category is only useful when NSURLSession is present
+#if defined(__IPHONE_7_0) || defined(__MAC_10_9)
+
+////////////////////////////////////////////////////////////////////////////////
+#pragma mark - NSURLRequest+CustomHTTPBody
+
+@interface NSURLRequest (CustomHTTPBody)
+/**
+ *   Unfortunately, when sending POST requests (with a body) using NSURLSession,
+ *   by the time the request arrives at OHHTTPStubs, the HTTPBody of the
+ *   NSURLRequest has been reset to nil.
+ *
+ *   You can use this method to retrieve the HTTPBody for testing and use it to 
+ *   conditionally stub your requests.
+ */
+- (NSData *)OHHTTPStubs_HTTPBody;
+@end
+
+#endif /* __IPHONE_7_0 || __MAC_10_9 */
--- a/OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.m
+++ b/OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.m
@@ -1,0 +1,86 @@
+/***********************************************************************************
+*
+* Copyright (c) 2016 Sebastian Hagedorn, Felix Lamouroux
+*
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
+* in the Software without restriction, including without limitation the rights
+* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the Software is
+* furnished to do so, subject to the following conditions:
+*
+* The above copyright notice and this permission notice shall be included in
+* all copies or substantial portions of the Software.
+*
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+* THE SOFTWARE.
+*
+***********************************************************************************/
+
+#import "NSMutableURLRequest+HTTPBodyTesting.h"
+
+#if defined(__IPHONE_7_0) || defined(__MAC_10_9)
+
+////////////////////////////////////////////////////////////////////////////////
+#pragma mark - Imports
+
+#import "OHHTTPStubsMethodSwizzling.h"
+
+////////////////////////////////////////////////////////////////////////////////
+#pragma mark - NSURLRequest+CustomHTTPBody
+
+NSString * const OHHTTPStubs_HTTPBodyKey = @"HTTPBody";
+
+@implementation NSURLRequest (CustomHTTPBody)
+
+- (NSData*)OHHTTPStubs_HTTPBody
+{
+    return [NSURLProtocol propertyForKey:OHHTTPStubs_HTTPBodyKey inRequest:self];
+}
+
+@end
+
+////////////////////////////////////////////////////////////////////////////////
+#pragma mark - NSMutableURLRequest+HTTPBodyTesting
+
+typedef void(*OHHHTTPStubsSetterIMP)(id, SEL, id);
+static OHHHTTPStubsSetterIMP orig_setHTTPBody;
+
+static void OHHTTPStubs_setHTTPBody(id self, SEL _cmd, NSData* HTTPBody)
+{
+    // store the http body via NSURLProtocol
+    if (HTTPBody) {
+        [NSURLProtocol setProperty:HTTPBody forKey:OHHTTPStubs_HTTPBodyKey inRequest:self];
+    } else {
+        // unfortunately resetting does not work properly as the NSURLSession also uses this to reset the property
+    }
+
+    orig_setHTTPBody(self, _cmd, HTTPBody);
+}
+
+/**
+ *   Swizzles setHTTPBody: in order to maintain a copy of the http body for later
+ *   reference and calls the original implementation.
+ *
+ *   @warning Should not be used in production, testing only.
+ */
+@interface NSMutableURLRequest (HTTPBodyTesting) @end
+
+@implementation NSMutableURLRequest (HTTPBodyTesting)
+
++ (void)load
+{
+    orig_setHTTPBody = (OHHHTTPStubsSetterIMP)OHHTTPStubsReplaceMethod(@selector(setHTTPBody:),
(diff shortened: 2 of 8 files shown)


OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.h
```diff
@@ -0,0 +1,46 @@
...
+@end
+
+@interface NSURLRequest (CustomHTTPBodyProperty)
+
+/**
+ *   Unfortunately when you send POST requests (with a body) using NSURLSession,
+ *   by the time the request arrives to OHHTTPStubs, the HTTPBody of the NSURLRequest
+ *   can't be accessed anymore.
+ *
+ *   You can use this method to retrieve the HTTPBody for testing and
```
shagedorn: `for testing and`...?

AliSoftware:
Cool, looks interesting !

Don't forget to add an entry in the CHANGELOG to credit yourself.

I'm going on vacation for two weeks so I won't have time to check that PR until I return.

OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.m
```diff
@@ -0,0 +1,89 @@
...
+        if (didAddMethod) {
+            class_replaceMethod(class,
+            swizzledSelector,
+            method_getImplementation(originalMethod),
+            method_getTypeEncoding(originalMethod));
+        } else {
+            method_exchangeImplementations(originalMethod, swizzledMethod);
+        }
+
+        NSLog(@"Enabled method swizzeling setHTTPBody: on NSMutableURLRequest");
```
AliSoftware: You should probably remove the NSLogs now that you've debugged its proper integration?

OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.m
```diff
@@ -0,0 +1,89 @@
...
+ *   Courtesy of http://nshipster.com/method-swizzling/
+ */
+
++ (void)load {
+    static dispatch_once_t onceToken;
+
+    dispatch_once(&onceToken, ^{
+        Class class = [self class];
+
+        SEL originalSelector = @selector(setHTTPBody:);
```
AliSoftware: Note that although this is a perfectly valid code for swizzling, I already use a different technique in [OHHTTPStubs+NSURLSessionConfiguration.m](https://github.com/AliSoftware/OHHTTPStubs/blob/master/OHHTTPStubs/Sources/NSURLSession/OHHTTPStubs%2BNSURLSessionConfiguration.m#L42-L58) to perform a similar swizzling.

Maybe it could be worth mutualizing the swizzling somehow (even if the one in `OHHTTPStubs+NSURLSessionConfiguration.m` is typed to return the specific signature of the `NSURLSession` methods I swizzle for now but you could change that and maybe move the swizzle helper code in a common file instead, idk)

Not sure which of the two implementations is better; I like the fact in mine that we store the original implementation in a `static` with an explicit name so we explicitly call `orig_defaultSessionConfiguration(…)` when we need to — instead of having to send the `OHHTTPStubs_setHTTPBody:` message to `self` inside the swizzled `OHHTTPStubs_setHTTPBody` implementation, which might be confusing even if you properly added a comment to explain it — but that's a matter of taste. But whatever solution we choose it would at least avoid code duplication.

OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.m
shagedorn: The idea was to make sure this never ends up in any code unnoticed. However, swizzling is used in other parts of `OHHTTPStubs`, too, and the framework should generally only be used for testing and the likes, so I agree, this can probably be removed.

OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.m
shagedorn: I agree that the static variable is easier to grasp than the commented message sending approach. I'll take a look how we can refactor this to reuse the swizzling code.

felixLam:
Thanks for getting back so quickly! Enjoy your vacation, we will improve this PR in the meantime.

shagedorn:
If merged, you'd probably want to...

- [ ]  Adjust the [Wiki entry](https://github.com/AliSoftware/OHHTTPStubs/wiki/Testing-for-the-request-body-in-your-stubs)
- [ ]  Post to issue [#52](https://github.com/AliSoftware/OHHTTPStubs/issues/52)
- [ ]  Adjust the version number in the changelog as appropriate

OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.m
```diff
@@ -0,0 +1,86 @@
...
+typedef void(*OHHHTTPStubsSetterIMP)(id, SEL, id);
+static OHHHTTPStubsSetterIMP orig_setHTTPBody;
+
+static void OHHTTPStubs_setHTTPBody(id self, SEL _cmd, NSData* HTTPBody)
+{
+    // store the http body via NSURLProtocol
+    if (HTTPBody) {
+        [NSURLProtocol setProperty:HTTPBody forKey:OHHTTPStubs_HTTPBodyKey inRequest:self];
+    } else {
+        // unfortunately resetting does not work properly as the NSURLSession also uses this to reset the property
```
AliSoftware: If setting it to `nil` doesn't work, wouldn't it at least be an acceptable workaround to set it ton an empty `NSData` instead?

OHHTTPStubs/Sources/NSURLSession/OHHTTPStubsMethodSwizzling.h
```diff
@@ -0,0 +1,51 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ ***********************************************************************************/
+
+////////////////////////////////////////////////////////////////////////////////
```
AliSoftware: This header file should not be part of the public API. You might want to alter the `podspec` accordingly [here](https://github.com/AliSoftware/OHHTTPStubs/blob/master/OHHTTPStubs.podspec#L55) to make it private (and thus make sure consumers of the pod won't be able to `#import` it from their app project)

AliSoftware:
Finally had time to check that PR :nada:

Please address the above comments, then rebase the branch on top of master — because as master has evolved since your PR, your branch is now in conflict (probably only because of the CHANGELOG entry, so not a complex conflict to fix :wink:)

Otherwise I'm 👍

AliSoftware:
>  * [ ] Adjust the version number in the changelog as appropriate

FYI, in general I use a special `## Master` entry for that (instead of the `5.x` you created), so that we can add more and more entries and only specify what the exact version number it should be (according to semver) when releasing a new version only :wink:

So when you'll be rebasing, you'll see that `## Master` entry is now already existing on `master` (since I already merged other PRs that had it), so just put your changelog entry below the others of that section and I'll take care of changing `## Master` to the proper version when I decide to do a release :wink:

OHHTTPStubs/Sources/NSURLSession/NSMutableURLRequest+HTTPBodyTesting.m
felixLam: I was hesitant as the session may rely on it being nil. Not sure if we can differentiate between the callers here. Also I believe that in production code it rarly happens that you set the body and then later reset it.

felixLam: ah. No the reason we can't do that is that the session calls this method with nil which lead to the initial problem. If we touch the protocol property here we lose the benefit of this PR.

AliSoftware: Oh I see, right.
Didn't understand your code comment at first but makes total sense now 👍

AliSoftware:
Ping @felixLam ?

* [ ] Remove `OHHTTPStubsMethodSwizzling.h` from the public headers in the `podspec`
* [ ] Rebase

shagedorn:
Sorry about the delay, we had a busy week and @felixLam was on holidays. We'll tackle it next week!

AliSoftware:
np, thx

felixLam:
@AliSoftware I made the requested changes and rebased on master

AliSoftware:
Thx  :+1:

_(I don't know what's wrong with travis not reporting the test results on this PR suddenly… will investigate and merge once ok)_

felixLam:
@AliSoftware there was a "minor" outage on GitHub today.

AliSoftware:
@felixLam Ah good point that was probably it. Tested locally and force-merged onto master, (let wait for travis to check master now just to be sure… :wink:)

AliSoftware:
@felixLam In an effort to try and follow [Moya's Contributor Guidelines](https://github.com/Moya/contributors), and to thank your for your recent contribution to `OHHTTPStubs`, I just added you as contributor to `OHHTTPStubs` so that you can now merge PRs and manage issues!
_(Note: master is locked against direct commits, in order to force us to make PRs pass tests before we can merge any new code to master)_

felixLam:
@AliSoftware Thank you very much. Credit where credit is due: @shagedorn did most of the heavy lifting here with the method swizzling.

AliSoftware:
Right, added @shagedorn as a contributor as well just now :wink:

mikelupo:
How should we #import the ability to utilize this new code? I can't seem to build it with just #include <OHHTTPStubs/OHHTTPStubs.h>

Note, I'm not using cocoapods to install. I'm dragging the xcodeproj into a frameworks group in my own xcode project. that's how I've always used OHHTTPStubs.

Thanks in advance

felixLam:
@mikelupo The should not require any changes in your end as far as I remember and should just magically work.

mikelupo:
Thanks for the quick reply. 
Looks like its my problem. It's mainly because of the way that I integrate. 
When I removed the existing OHHTTPStubs.xcodeproj from my workspace, it also removed references to it from my build phases where I was building/linking it into my bundle.  So of course it would not find it. 
:)


Now that I've resolved that, it still won't build unless I import:

`#import <OHHTTPStubs/NSMutableURLRequest+HTTPBodyTesting.h>`



In my test, I'm using the new method call like this:

`NSData *myData = [request OHHTTPStubs_HTTPBody];`


Mike



-----Original Message-----

AliSoftware:
@felixLam @shagedorn while doing the 5.1.0 release, I just realized that maybe I merged that PR too soon. I mean, it should work, but:

* [ ] It's missing Unit Tests. We should add at least one unit test for this feature, ensuring that the code works and that there won't be any regression later
* [ ] The file where you added the category is called `NSMutableURLRequest+HTTPBodyTesting.h/m` but is in fact an extension on `NSURLRequest`, not `NSMutableURLRequest`, so for consistency's sake we should probably rename it

That doesn't seem much work to do and make it right, do you think you'd be able to throw a quick PR to add that soon, so I can include it properly in the upcoming 5.1.0 release?

Thx!

felixLam:
Absolutely. Will look at this tomorrow.

AliSoftware:
Cool Thx 👍

AliSoftware:
ping @felixLam ?

shagedorn:
@mikelupo the `OHHTTPStubs.h` file only includes the bare minimum of imports, which I assume is intentional ( @AliSoftware ?). To get everything, import `OHHTTPStubsUmbrella.h`, or use the modular import using `@import OHHTTPStubs;`, which will use the umbrella header.

I'm working on the remaining issues (rename & tests) right now and will submit a PR later today.

AliSoftware:
> the OHHTTPStubs.h file only includes the bare minimum of imports, which I assume is intentional

That's sadly a legacy issue, because when I created `OHHTTPStubs` years ago, I didn't think it would grow that much, and I only had one public class, named `OHHTTPStubs`, which have the same name as the pod/framework itself.

Now that `OHHTTPStubs` has grown a lot, `OHHTTPStubs.h` remains the header of only the `OHHTTPStubs` main class (like every other `Foo` class has its dedicated `Foo.h` header).  
I can't use `OHHTTPStubs.h` itself as the umbrella header as well (because of cross-import reasons — I can't import every `.h` in `OHHTTPStubs.h` because some of those headers needs the `OHHTTPStubs` class to be defined first — and for subspecs reasons — especially, the headers we need to import depend on the subspecs people opt-in in their `Podfile`). I tried in some branches to improve that and make `OHHTTPStubs.h` the umbrella header but sadly it always led to complex problems and incompatibility with modular imports and all.

That's because of all these reasons that the umbrella header being `OHHTTPStubsUmbrella.h` and not `OHHTTPStubs.h`

But anyway, using the modular import `@import OHHTTPStubs;` is indeed the way to import modules nowadays, which thanks to the `.modulemap` I wrote automatically import the right umbrella header for you.

---

Looking fwd to your PR @shagedorn so that we can release the next version 👍

felixLam:
@AliSoftware Sorry for not following through with my promise on Friday. We had an unscheduled workload due to a rejection ;)

AliSoftware:
np I know those situations 😉

Issue #162, opened 2016-03-23
Artefacts built with Xcode prior to 7.3 do not work anymore


AlexDenisov:
He there,

Just found recently that artefact I downloaded from "Releases" page do not work with Xcode 7.3 because of Swift 2.2 and ABI compatibility.
I rebuilt the framework using new Xcode and issue has gone.
It would be nice if you can upload another build, e.g. something like `OHTTPStubs.framework-Swift-2.2.zip`.

Thank you.

AliSoftware:
Hello,

Thanks for letting me know.

Problem is, the artefact uploaded in the GitHub Releases is there for Carthage compatibility. And Carthage explicitly expects it to be named `NameOfTheLibrary.framework.zip` explicitly; so I don't see how I can make it work for Carthage for both Swift 2.1 and 2.2, as I will have to choose which one to name it so that Carthage should pick it, and that choice is impossible as it depends on what Swift version people are using.
So instead I think I'll have to release a new version… with no real new feature but just to allow Carthage users to pick one or the other :unamused: 

_That's one reason why I don't recommend using Carthage which rely on Swift ABI compatibility which we know will break, and encourage using other solutions like CocoaPods instead._

iamtomcat:
You could also not include a prebuilt framework. While it makes it slower for people as they would have to wait for carthage to build it, but it would fix the incompatibility issue.

AlexDenisov:
@AliSoftware I can't say what is better for Carthage, but newly built framework in any form will work for us, since we use custom solution for managing dependencies.

@iamtomcat I think it's still helpful to have both options.

AliSoftware:
@AlexDenisov unfortunately given how Carthage works we can't both propose the zip for Carthage to download and propose the option for Carthage to build from source. Either the zip is there either it isn't but we can't have multiple zips for Carthage to use and if there is a zip Carthage won't build from source. I know, right?

AliSoftware:
I think the only reasonable option here is what @iamtomcat suggests: for the next release (expected in a few days after merging the last recent submitted PRs) I won't include any zip at all so that Carthage will be forced to rebuild from source in order to be sure it's compatible with both Swift 21 and 2.2 and avoid ABI incompatibility.

AlexDenisov:
@AliSoftware TIL :)
The way you choose will work us as well.
Thank you for working on the problem.

iamtomcat:
Just as a workaround atm. I've added
[...]

PR #169, opened 2016-04-17
Let Travis auto deploy carthage build

thii:
This tells Travis to build and attach prebuilt framework archives to the GitHub releases.
In order for this to work you'll need to add your encrypted API key by running

    travis setup releases

See https://docs.travis-ci.com/user/deployment/releases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,16 @@ env:
 
 script:
   - rake "$RAKETASK"
+
+before_deploy:
+  - carthage build --no-skip-current
+  - carthage archive OHHTTPStubs
+
+deploy:
+  provider: releases
+  api_key:
+    secure:
+  file: OHHTTPStubs.framework.zip
+  skip_cleanup: true
+  on:
+    tags: true


AliSoftware:
Interesting! Does this properly only do that on tags, not on any other master commit, right?

---

FYI, the last releases of `OHHTTPStubs` stopped attaching prebuilt framework archives to the GH release because it led to problem for Carthage users. Indeed, some people were still using Xcode 7.2 (thus Swift 2.0 or 2.1) while other already migrated to Xcode 7.3 (thus Swift 2.2), and using a prebuilt framework led to ABI incompatibility issues (see #162 for more details)

That's why I preferred to let people build the framework with the same Xcode version that they were building their app, instead of providing a prebuilt-framework… that wouldn't work for half of the users.

Maybe I'm gonna restore the prebuilt-framework for the next release, considering that now people had time to migrate to Xcode 7.3, but I don't want to penalize people that are force for some reason to still use 7.2 and Swift 2.1…

AliSoftware:
Merged via 99feb7fc12b14e03e83147684dd3c409b4a247a3 (at last!)

Thanks for the tip!

AliSoftware:
Just released 5.1.0 with this travis config enabled and it worked and indeed released the Carthage build automagically 👍 Thanks

PR #172, opened 2016-05-05
Adjust parsing of Mocktail files to allow Headers to start on line 4

### LINKED ISSUE #170: Mocktail headers not parsed correctly?
Firstly, thanks so much for OHHTTPStubs.

I'm running into this error with loading stubs from Mocktail files:
`Request failed: unacceptable content-type: content-type: application/json`

From reading the code and this [test project](https://github.com/JinlianWang/TestHTTPStubs/blob/master/TestHTTPStubs/logos_ebay.tail#L4) it looks like Line 4 of Mocktail files is always interpreted as the value of the `Content-Type` header.

The documentation and reference implementation here https://github.com/puls/objc-mocktail/blob/master/Mocktail/Mocktail.m#L247 (which was transferred from the `square` github account to `puls`) shows that Line 4 and every following line until a blank line are interpreted header keys and values.

Fixing this might be considered an breaking change. But since the Mocktail format is far from standardized we could accept both formats by checking for a colon `:` on Line 4.

---
PS: Is HTTPMessage similar to Mocktail? I tried to find some documentation of it on the web but kept coming up with Java and Ruby documention of classes with that name 😆.

Ashton-W:
Fixes #170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # OHHTTPStubs — CHANGELOG
 
+* Adjusted parsing of Mocktail files to allow headers to start on line 4.  
+  [@Ashton-W](https://github.com/Ashton-W), [#172](https://github.com/AliSoftware/OHHTTPStubs/pull/172)
+
 ## [5.0.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/5.0.0)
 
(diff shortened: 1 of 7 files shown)


AliSoftware:
:ok_hand:

Could you add a line to credit yourself in the CHANGELOG?

PR #173, opened 2016-05-09
Update project to declare swift code

rodericj:
When installing from source one might run into
an issue suggesting that libswiftCore was not
found. This was problematic in Xcode 7.3 and
would cause all tests targets to fail.


AliSoftware:
Good call! Could you also add an entry in the `CHANGELOG` to credit yourself and link the PR? Thx!

rodericj:
Done.

PR #180, opened 2016-06-06
HTTP Body Tests

shagedorn:
This PR addresses two remaining issues from https://github.com/AliSoftware/OHHTTPStubs/pull/166.

* Renames file and category to improve consistency and avoid misunderstandings
* Adds two test cases:
    * using the new method, the http body should be accessibly
    * using the native method, the http body should not be accessible (else we didn't need the changes after all)

The renamed file isn't explicitly mentioned in the Podspec, so there's no change in that file. I assume users who have not enabled modules in their project/target will have to import `NSURLRequest+HTTPBodyTesting.h` (or the umbrella header) manually, or is there any other place to put it?


AliSoftware:
As `NSURLRequest+HTTPBodyTesting.h` is already in `OHHTTPStubsUmbrella.h` it seems, I don't think there is any other place to add it. It's mentioned in the wiki page as well so we should be good!

AliSoftware:
5.1.0 is out, including this change!

Issue #46, opened 2013-12-19
Stubs request tests pass, but stubs response block not called


sibljon:
Hi,

Here's my setup for the stubbing:

``` obj-c
id<OHHTTPStubsDescriptor> stub = [OHHTTPStubs stubRequestsPassingTest:^BOOL(NSURLRequest *request) {

    return [request.URL.absoluteString containsString:kVisitReviewPath];

} withStubResponse:^OHHTTPStubsResponse*(NSURLRequest *request) {

    return [OHHTTPStubsResponse responseWithFileAtPath:OHPathForFileInBundle(@"patient_visit_review.json",nil)
                                            statusCode:200
                                               headers:@{@"Content-Type":@"text/json"}];
}];
```

This test is being passed multiple times (I'm breaking on `firstStubPassingTestForRequest:` and `foundStub` always contains the exact `OHHTTPStubsDescriptor` I'm expecting); however, the stub response block never gets called.

Here's what I've verified so far, based on reading the Github Issue history of this repo:
1. `setEnabled:` gets called with `YES`, so `OHHTTPStubsProtocol` gets registered
2. I'm using AFNetworking 2.0, and `NSURLSession` is initialized with `[NSURLSessionConfiguration defaultSessionConfiguration]`
3. There is only one singleton instance of `OHHTTPStubs` being used

Does anyone have any idea where I can look next to keep debugging?

PR #182, opened 2016-06-20
Support for Swift 3.0 projects

djbriane:
This is just a very basic migration to support Swift 3.0 (there might be additional changes needed to conform to the updated API design guidelines). Posting it here in case you want to provide this in a branch for others who are updating their projects (or as a starting point for a full migration).

### DIFF
--- a/OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
+++ b/OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
@@ -810,11 +810,15 @@
 					09110A5019805F4800D175E4 = {
 						TestTargetID = 09110A4019805F4800D175E4;
 					};
+					093442DD1B80EC4A00A91535 = {
+						LastSwiftMigration = 0800;
+					};
 					095981D119806A7900807DBE = {
 						TestTargetID = 095981C119806A7900807DBE;
 					};
 					725CD99A1A9EB65100F84C8B = {
 						CreatedOnToolsVersion = 6.1.1;
+						LastSwiftMigration = 0800;
 					};
 					EA100AB61BE15BE400129352 = {
 						CreatedOnToolsVersion = 7.1;
@@ -1438,6 +1442,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = "com.alisoftware.${PRODUCT_NAME:rfc1034identifier}";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = 3.0;
 				WRAPPER_EXTENSION = xctest;
 			};
 			name = Debug;
@@ -1464,6 +1469,7 @@
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = "com.alisoftware.${PRODUCT_NAME:rfc1034identifier}";
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SWIFT_VERSION = 3.0;
 				WRAPPER_EXTENSION = xctest;
 			};
 			name = Release;
@@ -1605,6 +1611,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = "com.alisoftware.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = OHHTTPStubs;
 				SKIP_INSTALL = YES;
+				SWIFT_VERSION = 3.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Debug;
@@ -1628,6 +1635,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = "com.alisoftware.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = OHHTTPStubs;
 				SKIP_INSTALL = YES;
+				SWIFT_VERSION = 3.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Release;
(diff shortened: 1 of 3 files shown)


AliSoftware:
Thanks for the Pull Request! 👍 

I can't merge it on `master` though, as it's way too early to migrate the `master` branch of `OHHTTPStubs` to Swift 3 — while many people will continue using Xcode 7.3 until 8.0 is out of beta, and many people will continue using Swift 2.3 by then anyway.  
But I'll of course consider creating a dedicated `Swift3` branch to merge that PR into, so that people could point to it :+1:

----

In the meantime, it would be super-cool if you could **wrap the changes you made in the source code in some `#if swift(>=3.0)` blocks**, so that the source code would still be compatible with both Swift 2.x and Swift 3.0.

This would allow the pod to be installable and compilable on whatever Swift version people are using (at least if they use CocoaPods, as the podspec doesn't rely on the Xcode project but only on the source files) without the need to manage multiple branches (I could merge _that_ to `master` :wink:)

djbriane:
Agreed - the most logical approach for now would be to create a `swift3` branch so any others who want to migrate their projects can point to that branch ([Alamofire took this approach](https://github.com/Alamofire/Alamofire/issues/1316)).

In regards to your suggestion to add conditionals, I spent some time working through those changes and came to the conclusion I don't think it is the best approach for the following reasons:

1. Several of the method signatures need to be modified in order to conform to proposal [#0046](https://github.com/apple/swift-evolution/blob/master/proposals/0046-first-label.md) regarding consistent label behavior for methods and functions. This results in the following block of code being replicated several times and obviously introduces a lot of code duplication:

  ```Swift
  #if swift(>=3.0)
      public func fixture(_ filePath: String, status: Int32 = 200, headers: [NSObject: AnyObject]?) -> OHHTTPStubsResponse {
          return OHHTTPStubsResponse(fileAtPath: filePath, statusCode: status, headers: headers)
      }
  #else
      public func fixture(filePath: String, status: Int32 = 200, headers: [NSObject: AnyObject]?) -> OHHTTPStubsResponse {
          return OHHTTPStubsResponse(fileAtPath: filePath, statusCode: status, headers: headers)
      }
  #endif
  ```

2. Swift 3 introduces new API Design Guidelines (proposal [#0023](https://github.com/apple/swift-evolution/blob/master/proposals/0023-api-guidelines.md)) which resulted in numerous updates to the standard library (proposal [#0006](https://github.com/apple/swift-evolution/blob/master/proposals/0006-apply-api-guidelines-to-the-standard-library.md)). There are many uses of these libraries in the project, which results in variations of the following code replicated throughout the file:

  ```Swift
  public func isScheme(_ scheme: String) -> OHHTTPStubsTestBlock {
      #if swift(>=3.0)
          return { req in req.url?.scheme == scheme }
      #else
          return { req in req.URL?.scheme == scheme }
      #endif
  }
  ```

Rather than force this much code duplication into the project, it seems best for now to keep this in a branch until it can be merged back into master once Swift 3 gets mainstream adoption. Keep in mind I'm no expert with Swift so if anyone has a more elgant approach I'd love to learn about it.

Thoughts?

AliSoftware:
I think we could do the migration in multiple steps:

1. create a new (major) version of OHHTTPStubs that alter its API in order to follow the Swift 3 API Guidelines, but still be compatible/compiled with Swift 2. This means keep using Swift 2, but simply alter the method signatures of the pod's API to prepare it for the migration, and release a new major version so people prepare to this new API even in Swift 2
2. from that version, alter the implementation where needs be (potentially using the `#if` tests if its possible not to have too much code duplication) so we can have a Swift 3 branch

In summary, address points 1 & 2 separately in order to help some progressive migration.
Do you think that would be worth it and help people prepare for migration?

mxcl:
As a suggestion for maintaining both 2.x and 3.x, at least as far as API renames, you could make a compatibility layer:

```swift
#if swift(>=3.0)
#else
extension NSURLRequest {
   var url: NSURL { return URL }
}
#endif
```

To keep things neat this can be all kept in a separate file.

Not that I think you *need* to maintain both, I don't intend to personally for my projects.

AliSoftware:
Thx @mxcl for the suggestion (wow THE Max Howell commenting on my repo :blush: ^^) this might indeed reduce code duplication during the migration phase :+1:

_(To be clear I don't intent on maintaining Swift 2 compatibility in the long run; that would only be for the transition phase, until ~ end of 2016 I guess)_

mxcl:
I’m using the following in my fork so that OHHTTPSubs compiles against Swift 2.2, 2.3 & 3.0.
```swift
#if swift(>=3.0)
#else
    extension OHHTTPStubs {
        private class func stubRequests(passingTest passingTest: OHHTTPStubsTestBlock, withStubResponse: OHHTTPStubsResponseBlock) -> OHHTTPStubsDescriptor {
            return stubRequestsPassingTest(passingTest, withStubResponse: withStubResponse)
        }
    }

    extension NSURLRequest {
        private var httpMethod: String? { return HTTPMethod }
        private var url: NSURL? { return URL }
    }

    extension NSURLComponents {
        private convenience init?(url: NSURL, resolvingAgainstBaseURL: Bool) {
            self.init(URL: url, resolvingAgainstBaseURL: resolvingAgainstBaseURL)
        }
    }

    private typealias URLRequest = NSURLRequest

    extension URLRequest {
        private func value(forHTTPHeaderField key: String) -> String? {
            return valueForHTTPHeaderField(key)
        }
    }
#endif
```

AliSoftware:
@djbriane Would you be OK if I close your PR in favor of #192 which seems more advanced as it uses @mxcl 's suggestions and will be able to support both Swift 2.2, 2.3 and 3.0 in the same branch (if I understood #192 correctly)?

Thanks for the efforts anyway, and sorry it took me so much time to give some feedback of this Swift 3 support!

djbriane:
@AliSoftware Sure, not a problem. I'll go ahead and close it.

AliSoftware:
Thx and thanks again for your efforts on this PR!

PR #184, opened 2016-06-21
Support Swift 2.3

ikesyo:
This should be compatible with both Swift 2.2/Xcode 7.3 and Swift 2.3/Xcode 8.

- [x] Depends on #185 
- [x] Wait for the next Xcode 8 beta to remove a weird workaround of 3c3be9d
- [x] Wait for Xcode 8 beta 2 on Travis


AliSoftware:
Thanks for the PR !

Wondering if I should really merge the last commit of your PR 80860843e066450b95ab09bb01960e5f2949c11d (the one containing that patch described in Quick/Nimble#302) or better wait for the next Xcode beta to see if it's fixed. Keeping that patch in the long run in the repo would seem hacky to me (because let's be honest, if I merge this patch commit I'll forget about it in a few weeks/months and never remove it later :wink:)

ikesyo:
Thanks for your thought! I'm OK to wait for the next Xcode beta to remove the workaround. In the meantime I'll using the branch personally.

AliSoftware:
Hey @ikesyo !

It seems like the workaround is no longer needed in Xcode 8 beta 2 ? https://github.com/Quick/Nimble/pull/316

Could you thus please remove the workaround commit from this PR so we can move forward and merge? Thx!

AliSoftware:
Note that we might also want to tweak [the `.travis.yml` file](https://github.com/AliSoftware/OHHTTPStubs/blob/master/.travis.yml) by adding a build matrix so the CI check the build both with Xcode 7.3 and Xcode 8 (it is currently only using the `osx_image: xcode7.3`.

You can find some inspiration [on this `.travis.yml` file on my other library](https://github.com/AliSoftware/Reusable/blob/master/.travis.yml) where I already did that kind of configuration.

ikesyo:
Travis now supports Xcode 8 beta 2: https://twitter.com/travisci/status/752704816116293632. But the iOS tests are randomly failed as seen in the following:

- https://travis-ci.org/AliSoftware/OHHTTPStubs/builds/144138180
- https://travis-ci.org/AliSoftware/OHHTTPStubs/builds/144140405
- https://travis-ci.org/AliSoftware/OHHTTPStubs/builds/144144290

> ▸ Signing /Users/travis/Library/Developer/Xcode/DerivedData/OHHTTPStubs-gozjzzlydcppgdbvdohewsnafzwl/Build/Products/Debug-iphonesimulator/OHHTTPStubs\ iOS\ Fmk\ Tests.xctest
> ** TEST FAILED **
> rake aborted!
> Command failed with status (65): [set -o pipefail && xcodebuild -workspace O...]

AliSoftware:
Sadly that happens more and more on Travis lately, builds taking ages to start and builds randomly failing, while they pass perfectly locally.
Given that we now have 8 parallel builds for each PR now (because of the build matrix testing the 4 targets, each in Xcode 7 and 8) there are more chances that one of those 8 randomly fail because of that.

Not sure if I can do anything about it. I could deactivate my timing tests for all but one build, that would probably help as those expect the response to be in a specific time window to ensure the given responseTime is respected… but if the Travis VM is too solicited and slow that doesn't help. But the most build fails I've seen lately are not because of them but more because of Travis failing to launch th Xcode test command, so I'm puzzled…

I just relaunched the Travis build that failed and hope it will pass this time, until I have more time to investigate that Travis bug.

PR #195, opened 2016-08-19
Add info about OHHTTPStub/Swift to readme

ianks:


AliSoftware:
@ianks Thanks a lot for this PR. That's a very good idea to add this information.

However, the added information is not totally true (that proves that adding that information is indeed needed 😉 ), because the `OHHTTPStubs/Swift` subspec [only depends on `OHHTTPStubs/Core`](https://github.com/AliSoftware/OHHTTPStubs/blob/master/OHHTTPStubs.podspec#L88) (so that you can opt-in only for what you need), and does not automatically brings everything in the default `OHHTTPStubs/Default` subspec.

So in practice, if you *only* put `pod 'OHHTTPStubs/Swift` in your `Podfile`, you'll only have the Core + Swift features, but you won't have support for `NSURLSession`, `JSON` and `OHPathHelper` that is [provided by the default subspec](https://github.com/AliSoftware/OHHTTPStubs/blob/master/OHHTTPStubs.podspec#L41-L43). That's intended, so that people only wanting to opt-in for some subspecs but not all can configure to their liking.

So I think the best suggestion to give in the README to cover most of what people uses is to list both the default and the Swift subspecs:

```
pod 'OHHTTPStubs'
pod 'OHHTTPStubs/Swift'
```

AliSoftware:
I've rearranged some stuff in the README, about Swift 2.2/2.3/3.0 and about subspecs.

I've directly integrated your recommendations in that rework, explaining about the `Swift` subspec too. Feel free to review the last changes in the README on master to confirm it looks ok and more clear to you!

I'm thus closing this PR now. Thanks a lot for that suggestion!

PR #201, opened 2016-09-07
Swift 3.0: Fix Xcode 8 GM compatibility

ikesyo:


mxcl:
We want to keep the single `#if` block, this was a design decision in previous PRs, could you amend the patch conforming to the previous style? Thanks.

ikesyo:
Rebased

mxcl:
Many thanks.

AliSoftware:
:+1: thanks for the taking care of this fix 👌

AliSoftware:
5.2.1 is out, with this fix included 🎉 

* I cherry-picked this fix from `swift3` in `master`, so that people using CocoaPods can still point to `pod 'OHHTTPStubs, '~> 5.2'` and get that change too if they're using Xcode 8 GM
* I rebased the `swift3` branch on top of `master` so that the only commit making `swift3` differ from `master` is still `SWIFT_VERSION=3.0` only, as this branch's only purpose is for Carthage users only

For Swift 2.2 / 2.3 users, this 5.2.1 doesn't have much change compared to 5.2.0 (except README & docs typos) but I prefer keeping all my versions in sync to avoid confusions.

PR #205, opened 2016-09-14
Migrate samples to Swift 3

dhardiman:
Carthage seems to find this example project and fall over when the dependencies are checked using Xcode 8 as it’s not specifying the swift version. This appears to fix that. It does seem to build the framework target twice due to the cocoa pods usage for the example, but I don’t think that causes a problem. Perhaps in future the example project could link directly to the project rather than using CocoaPods to solve that, but probably not urgent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # OHHTTPStubs — CHANGELOG
 
+* Migrate samples to Swift 3
+  [@dhardiman](https://gitub.com/dhardiman)
+
 ## [5.2.1](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/5.2.0)
 
 * Fix typos in README and documentation.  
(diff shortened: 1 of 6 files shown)


AliSoftware:
👌 Could you also add an entry in the CHANGELOG to credit yourself for this migration? Thx!

dhardiman:
Sure, updated. Hope the format is ok above the previous release? Let me know if not and I'll fix.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # OHHTTPStubs — CHANGELOG
 
+* Migrate samples to Swift 3
```
AliSoftware: Could you add a full stop and 2 spaces at the end of that line so that the markdown formatting matches the other notes?

dhardiman: Updated

AliSoftware:
Travis-CI builds are currently failing due to a bug in the latest version of `xcpretty`: https://github.com/supermarin/xcpretty/issues/241 — Waiting for the fix.

AliSoftware:
Merged manually via ca8e9c295b859215b1ab8a9df8e0546a24e768e5

Thanks again!

PR #226, opened 2016-11-12
Minor documentation update for Swift 3.

manicaesar:
Previous documentation was not quite working in Swift, due to:
- String not implicitly casted to NSString / NSObject / AnyObject
- missing external parameter name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,8 @@ This example is using the Swift helpers found in `OHHTTPStubsSwift.swift` provid
 ```swift
 stub(isHost("mywebservice.com")) { _ in
   // Stub it with our "wsresponse.json" stub file (which is in same bundle as self)
-  let stubPath = OHPathForFile("wsresponse.json", self.dynamicType)
-  return fixture(stubPath!, headers: ["Content-Type":"application/json"])
+  let stubPath = OHPathForFile("wsresponse.json", type(of: self))
+  return fixture(filePath: stubPath!, headers: ["Content-Type":"application/json"])
 }
 ```
 


manicaesar:
> If it really is, maybe we should use NS_REFINED_FOR_SWIFT or similar in the definition to avoid this at call site?

Yes, maybe that would be the best option. But I think it should not be done within this PR - rather within some separate one :)

BTW: What's wrong with the travis?

AliSoftware:
If you see a "can't find module OHHTTPStubs" error in Travis I'm afraid that's probably my fault, a recent commit on master to try and disable timing tests on the CI (because they are not deterministic and depend too much on the machine executing them) seen to have broken something. Haven't had time to investigate further yet ☹️

manicaesar:
Yes, that's the case, I think changes in README.md should typically not affect builds :) Ok, I will wait until the problem is solved (I don't have too much spare time right now either). Fingers crossed :)

AliSoftware:
:+1:

AliSoftware:
The build has been fixed on `master` 🎉  You'll now be able to rebase on top of master.

However, I'm still surprised that the cast to `NSString` is necessary.

In the sample project we don't use it
* [Swift 2.3 version of the Sample project on master](https://github.com/AliSoftware/OHHTTPStubs/blob/master/Examples/Swift/MainViewController.swift#L76-L77)
* [Swift 3.0 version of the Sample project on the `swift3` branch](https://github.com/AliSoftware/OHHTTPStubs/blob/swift-3.0/Examples/Swift/MainViewController.swift#L76-L77)

So I think the only change necessary in the documentation should be the `self.dynamicType` -> `type(of:self)`. Could you update your PR accordingly?

Thx

manicaesar:
Great news :) 

However, I haven't invented the problem - this keeps emerging (Xcode 8.1).
screen shot 2016-11-14 at 20 20 41

I will check with your sample project. Maybe some compiler warning is disabled.

manicaesar:
Ok, that was simpler than I expected. In my project I used OHHTTPStubs from master branch, where fixture is defined as follows:
`public func fixture(filePath: String, status: Int32 = 200, headers: [NSObject: AnyObject]?) -> OHHTTPStubsResponse`
while on swift-3.0 branch it is defined that way:
`public func fixture(filePath: String, status: Int32 = 200, headers: [AnyHashable: Any]?) -> OHHTTPStubsResponse`
which does not cause any problem.
There is no signal in README.md that I should use swift-3.0 branch when using the pod in application that uses Swift 3.0. Maybe we should add one? Or maybe you are about to merge branch swift-3.0 to master soon? :)

AliSoftware:
I plan to wait a bit more before merging Swift-3 branch in master.

`master` is supposed to be compatible with 2.2, 2.3 and 3.0 (only emitting warnings for extraneous `_` parameters in Swift 2.x), at least that's the intent until Swift 3 is merged to master when Swift 3 adoption is wider.
So the only difference in the pod between `master` and `swift-3.0` branches are supposed to be the `SWIFT_VERSION` build setting (and the demo project being updated to Swift 3)
If that's not the case and there are more differences in there pod between those two branches, making `master` *not* Swift-3 compatible then imho we should add an `#if Swift(>=3.0)` to make it so

manicaesar:
Sorry for the delay. Ok, I finally squashed and rebased my PR to contain only required changes. Please check. Thank you for cooperation!

PR #238, opened 2017-03-28
Convert syntax to Swift 3.1.

cohilla:


Liquidsoul:
Hi @cohilla,

Thank you for your contribution!

I did not test the library on Xcode 8.3 yet but I am wondering if you've looked at the [`swift-3.0`](https://github.com/AliSoftware/OHHTTPStubs/tree/swift-3.0) branch?
I suppose that it would be a better starting point than `master` if we want to make the switch from major 2.x/minor 3.x to major 3.x/minor 2.x

Issue #233, opened 2017-01-23
Please set master to swift 3 :D


gperdomor:
Any reason why we master in still in swift 2.3?

Liquidsoul:
Hi @gperdomor,

I suppose that you are talking about the Carthage configuration?

The rationale behind the current setup can be found in the comments of #192

Moving to swift 3 would require to keep a swift 2.3 compatibility anyway as the swift version is not enforced "yet". It should be soon when Xcode will drop the support for it.
Because of that, maybe it is better to keep things as-is for now and drop support of 2.3 altogether once Xcode stops supporting it.

Are you having a specific issue with this current setup? 😊

piercifani:
Hello @Liquidsoul,

Yes, there is an issue here when you integrate this library using Carthage, since Carthage will use the project's `xcodeproj` to create a framework, and currently the project is set to compile with Swift 2.3 [check it here](https://github.com/AliSoftware/OHHTTPStubs/blob/master/OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj#L1322).

In order to get this working with Carthage you'd have to fork it and change this value to `3.0`which is less than ideal.

Since Swift 2.3 is closer and closer to deprecation (Xcode 8.3 release could be less than a month away) I think it's safe to set Swift 3 as the default for this project.

Thanks!

Liquidsoul:
Hi @piercifani,

We made things so you should not need to fork the repository to use Carthage and Swift 3.0.
As stated in the carthage part of [this section](https://github.com/AliSoftware/OHHTTPStubs/tree/swift-3.0#using-the-right-swift-version-of-ohhttpstubs-for-your-project), all you need to do is specify the `swift-3.0` branch in your `Cartfile` and you should be good to go 😃 

As Xcode 8.3 is coming soon, maybe it is a good time to work on the Swift 2.3 to Swift 3.0 migration of the project.

Liquidsoul:
I forgot that you can also use dedicated releases for the library which have the `-swift3` suffix like [here](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/5.2.3-swift3) for the latest one, and previous ones are available too!

piercifani:
oops... sorry, I didn't catch that! Thanks a lot!

AliSoftware:
[...]

PR #240, opened 2017-03-29
Xcode 8.3 and Swift 3.x support

Liquidsoul:
This pull request is mostly a merge of the `swift-3.0` branch to implement #233 :
- it makes `master` use Swift 3.0 as default.
- it fixes some string warnings in some logs from Xcode 8.3.

Once this is merged we will create a `swift-2.3` branch to allow users to use the legacy Swift version if they need it.

❓~One thing I am wondering is if the Travis CI configuration should not be updated in this PR to test both Xcode 8.3 and Xcode 8.2?
Maybe at least change the `osx_image` to `xcode8.3` instead of `xcode8` which from [travis' documentation](https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version) targets some Xcode 8 GM~ 🙄 
The build matrix has been updated to use Xcode 8.3.

A language badge has been added to the [`README`](https://github.com/AliSoftware/OHHTTPStubs/blob/000b57e27adc772ed7d13fb5e3319c2f72d9ca3b/README.md) file. [Language: Swift 3.1](https://swift.org)

### DIFF
--- a/Examples/Swift/OHHTTPStubsDemo.xcodeproj/project.pbxproj
+++ b/Examples/Swift/OHHTTPStubsDemo.xcodeproj/project.pbxproj
@@ -293,7 +293,7 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
-				SWIFT_VERSION = 2.3;
+				SWIFT_VERSION = 3.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Debug;
@@ -334,7 +334,7 @@
 				MTL_ENABLE_DEBUG_INFO = NO;
 				SDKROOT = iphoneos;
 				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
-				SWIFT_VERSION = 2.3;
+				SWIFT_VERSION = 3.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
 				VALIDATE_PRODUCT = YES;
 			};
@@ -351,7 +351,6 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = "com.alisoftware.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				SWIFT_VERSION = 3.0;
 			};
 			name = Debug;
 		};
@@ -366,7 +365,6 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = "com.alisoftware.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				SWIFT_VERSION = 3.0;
 			};
 			name = Release;
 		};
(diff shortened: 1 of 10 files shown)


AliSoftware:
I think by now it would be better to have the following branch setup:

* Make `master` always use the latest Swift version, even if it's early in adoption (so in our case today, Swift 3.1)
* Make dedicated branches for older versions, so `swift-2.3` and `swift-3.0`

Because if today we make `master` use 3.0 and make a `swift-2.3` and `swift-3.1` branch we'll later again fall back in the similar issue we have today once 3.1 will be widely adopted.

Thoughts?

AliSoftware:
> ❓One thing I am wondering is if the Travis CI configuration should not be updated in this PR to test both Xcode 8.3 and Xcode 8.2?

Yeah probably :+1:

_(Just to be clear, we won't need a build matrix to always build master on all Xcode's, just each branch — `master`, `swift-2.3` and `swift-3.0` will each have their `travis.yml` configured with the matching `osx_image` for the branch)_

Liquidsoul:
Is there really a difference between Swift 3.1 and Swift 3.0?
Because it does not seem to be the case or maybe I am missing something 🤔

AliSoftware:
I don't think there is any code change needed at all. 
The only requested changes is for Carthage users, again… who will require us to do a branch on which the _only_ difference compared to 3.0 would be the value of `SWIFT_VERSION`

AliSoftware:
Side note: we need to migrate OHHTTPStubs to Circle-CI some day… travis is still taking forever :'(

Examples/Swift/OHHTTPStubsDemo.xcodeproj/project.pbxproj
```diff
@@ -328,6 +333,7 @@
 				IPHONEOS_DEPLOYMENT_TARGET = 8.3;
 				MTL_ENABLE_DEBUG_INFO = NO;
 				SDKROOT = iphoneos;
+				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
 				SWIFT_VERSION = 2.3;
```
AliSoftware: Why is this line still 2.3? (Idea reading that from the GitHub web page on mobile, so hard to easily get the context of that line in the whole xcodeproj to check if it's normal, but seems strange to me)

Liquidsoul: This is the project level settings which is overwritten by the target level settings. I'll change that and remove the target level one.

AliSoftware:
@Liquidsoul Now that this is merged:

* `master` is in Swift 3.1
* I've moved the `swift-3.0` branch to be equal to `master`, so that people referencing that branch (especially people using Carthage) can still use it. We'll remove it as soon as we publish a new release/tag
* I've created a `swift-2.3` branch (that is equal to where `master` was before the merge of that PR)

Next steps is to:

* Add a commit to that new `swift-2.3` branch to add the shields.io badge in the README
* Create a new release
* Remove the `swift-3.0` branch
* Profit!

Liquidsoul:
@AliSoftware I've just pushed the badge in branch [`swift2.3`](https://github.com/AliSoftware/OHHTTPStubs/tree/swift-2.3).

AliSoftware:
👌

krunk4ever:
I noticed that the releases are versioned as:

- 5.2.2
- 5.2.2-swift3
- 5.2.3
- 5.2.3-swift3

Moving forward, will this be switched to?

- 5.2.4
- 5.2.4-swift23

AliSoftware:
Yep exactly that's what I plan indeed. Do you think we should do otherwise? Suggestions welcome. 

Also wondering if I should bump the next release to 6.0.0 because the switch from 2.3 to 3.0 is source-breaking (I think, will have to compare our APIs for the two versions to check if the call site has to be changed or not 🤔)…

krunk4ever:
I agree you should make 5.2.3 not compatible with the next version since they are not compatible, so 👍  on 6.0.0

Issue #222, opened 2016-10-31
ohhttpStubs_HTTPBody() not available for swift 3.0


isimpson:
Currently no way to access the stubbed request body in swift 3.0 without first casting the request to NSURLRequest.
This works:
```
let req: NSURLRequest = request as NSURLRequest
let body = req.ohhttpStubs_HTTPBody()
```
This does not:
```
let body = request.ohhttpStubs_HTTPBody()
```

There was an attempt to fix this with PR: https://github.com/AliSoftware/OHHTTPStubs/pull/216
It seems that this was reverted as it was never working correctly.

AliSoftware:
Yes it's a known issue without an easy solution that would work for _every_ subspec configuration so not sure we're gonna fix that, given how complex it could be to make it work for people using the subspec… while still not make the compilation failed for people not using it.

In the meantime the fix is easy: just cast your `request as NSURLRequest`. Not as pretty but easy enough and did the job until we find a configuration working for everybody without breaking compilation for others.

isimpson:
Ok, thanks for the quick response. May be nice to leave this issue to help others with the current workaround:
`(request as NSURLRequest).ohhttpStubs_HTTPBody()`

AliSoftware:
@isimpson Release 6.0.0 (#242) will finally fix this by finally making the Swift subspec always include NSURLSession so we were able to add the `hasBody()` matcher to the Swift API too!

PR #242, opened 2017-04-02
Release/6.0.0

### LINKED ISSUE #228: Proposal for Swift: hasBody OHHTTPStubsTestBlock
Hi there,

I'm not sure if there was the same idea already or not. But what do you think about adding new  `OHHTTPStubsTestBlock`.

```swift
public func hasBody(_ body: Data) -> OHHTTPStubsTestBlock {
    return { req in req.OHHTTPStubs_HTTPBody? == body }
}
```

I see the problem here, that the `Swift` subspec doesn't contain the `URLSession` subspec by default, which makes sense. But maybe it could a new one `SwftURLSession` or something like that?

In the end, I find this new test block very useful. And I'd love to add this in PR if the feature is approved 🚀

AliSoftware:
### Checklist

### Description

This branch contains the next release 6.0.0 of OHHTTPStubs.
This also closes #228 & #222.

### Motivation and Context

This PR is to ensure the 6.0.0 release branch compiles correctly in all environments, by letting the CI matrix check them all, before officially making the release.


AliSoftware:
@Liquidsoul 👌 

I've added one last commit to add more information for Carthage users. I think that:

* given most people have migrated to Swift 3.0 already
* and that people still using Swift 2.3 won't be in a hurry migrating to OHHTTPStubs 6.0.0 if they haven't migrated to Swift 3.0 yet and those people will still use 5.2.3
* it's time-consuming to maintain a specific "swift-2.3" branch _just_ for those users (which are likely not many, as projects in Swift 2.3 and not migrated to Swift 3.0 yet is rare by now)

Then I've decided to stop doing swift-2.3 specific branches _just_ for so few people. Those people have alternatives anyway, and I don't want to lose time maintaining that old branch any longer for an obsoleted Swift version which is already not supported anymore in latest Xcode.

README.md
```diff
@@ -3,9 +3,9 @@ OHHTTPStubs
 
 [![Platform](http://cocoapod-badges.herokuapp.com/p/OHHTTPStubs/badge.png)](http://cocoadocs.org/docsets/OHHTTPStubs)
 [![Version](http://cocoapod-badges.herokuapp.com/v/OHHTTPStubs/badge.png)](http://cocoadocs.org/docsets/OHHTTPStubs)
-[![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
+[![Carthage Swift 3.0/3.1](https://img.shields.io/badge/Carthage-Swift%203.x-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
 [![Build Status](https://travis-ci.org/AliSoftware/OHHTTPStubs.svg?branch=master)](https://travis-ci.org/AliSoftware/OHHTTPStubs)
-[![Language: Swift 3.1](https://img.shields.io/badge/Swift-3.1-orange.svg)](https://swift.org)
+[![Language: Swift 3.1](https://img.shields.io/badge/Swift-2.3%2F3.0%2F3.1-orange.svg)](https://swift.org)
```
Liquidsoul: Shouldn't the markdown badge `[Language: Swift 3.1]`, even if not displayed, be updated to what the shield shows for consistency? (meaning Swift-2.3/3.0/3.1)

AliSoftware: Indeed!

PR #245, opened 2017-05-04
Replace InfiniteLoop.dk URL with archive.org URL

iCarambaa:
The link to the article seems to be offline. I replaced it with a cached version on archive.org.


AliSoftware:
Thanks a lot for this PR! Well spotted!

Issue #247, opened 2017-05-09
Stub not working as expected


mrjimoy:
I tried to initialize unit tests on my project but seems that the stubs not being called. Each time I tried to debug, it always pass the stub's line. Anyone have an idea? Thanks.

**HomeConnectorTests.swift**
```swift
import XCTest
import OHHTTPStubs
import Nimble
@testable import Pods_Mobile_iOS

class HomeConnectorTests: XCTestCase {
    
    override func setUp() {
        super.setUp()

        OHHTTPStubs.setEnabled(true)
        
        OHHTTPStubs.onStubActivation() { request, descriptor, response in
            // ...
        }
    }
    
    override func tearDown() {
        OHHTTPStubs.removeAllStubs()

        super.tearDown()
    }
    
    func testGetHomeData_RequestsDefault_Return200() {
        TestsHelper.stubNetwork(forConnector: HomeConnector.self,
                                   endpoint: "home",
                                   stubFileName: "GET_Home_200")
        
        
        HomeConnector.sharedInstance.getHomeConfig(onSuccess: { (data, object) in
            // --> Not even entering this part
        }) { (data, error) in
            XCTAssertFalse(true)
        }
        
        waitForExpectations(timeout: 10, handler: nil)
    }
    
}
```
**TestsHelper.swift**
```swift
import UIKit
@testable import Pods_Mobile_iOS
import OHHTTPStubs

class TestsHelper: NSObject {
    
    static func stubNetwork(forConnector connector: NSObject.Type, endpoint: String, stubFileName: String) {
        stub(condition: isHost(HttpConnector.sharedFCHTTPClient().getKAPIBaseUrl()) && isPath(endpoint)) { request in
            // --> Not even entering this part also :(
            guard let path = OHPathForFileInBundle(stubFileName.appending(".json"), Bundle.main)
                else {
                    preconditionFailure("Stub not found!")
            }
            
            return OHHTTPStubsResponse(
                fileAtPath: OHPathForFile(path, type(of: self) as! AnyClass)!,
                statusCode: 200,
                headers: ["Content-Type":"application/json"]
            )
        }
    }
    
}
```

This is the connector I use to get the response from the API : 

**HomeConnector.swift**
```swift
class HomeConnector: MainApiConnector {

    static let sharedInstance = HomeConnector()
    
    fileprivate override init() {
        super.init()
    }
    
[...]

PR #248, opened 2017-05-12
Add precondition assertions in scheme, host matchers

Liquidsoul:
Also added an explanation of which part it is in an example URL.

### Checklist

### Description

Add preconditions so that user are better informed of why their requests are not stubbed.
This also adds information of what is a scheme, host or path in the documentation of each matcher.

### Motivation and Context

This will help users see why their matchers setup does not work (see #247).

### DIFF
--- a/OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift
+++ b/OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift
@@ -154,29 +154,38 @@ public func isMethodDELETE() -> OHHTTPStubsTestBlock {
 /**
  * Matcher for testing an `NSURLRequest`'s **scheme**.
  *
+ * e.g. the scheme part is `https` in `https://api.example.com/signin`
+ *
  * - Parameter scheme: The scheme to match
  *
  * - Returns: a matcher (OHHTTPStubsTestBlock) that succeeds only if the request
  *            has the given scheme
  */
 public func isScheme(_ scheme: String) -> OHHTTPStubsTestBlock {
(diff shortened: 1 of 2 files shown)


OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift
```diff
@@ -154,30 +154,39 @@ public func isMethodDELETE() -> OHHTTPStubsTestBlock {
...
  *
+ * e.g. the scheme part is `https` in `https://api.example.com/signin`
+ *
  * - Parameter scheme: The scheme to match
  *
  * - Returns: a matcher (OHHTTPStubsTestBlock) that succeeds only if the request
  *            has the given scheme
  */
 public func isScheme(_ scheme: String) -> OHHTTPStubsTestBlock {
+  precondition(!scheme.contains("://"), "The scheme part of an URL never contains '://'. Only use strings like 'https' for this value, and not things like 'https://'")
```
AliSoftware: Seems like `contains` was not available in Swift 2.3 🙄 Maybe add an `extension String` in the `#if !swift(>=3.0)` block at the top of this file to add support for it? https://github.com/AliSoftware/OHHTTPStubs/blob/5956336837ccb8d483119fede8cdee6590be8485/OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift#L30-L55

Liquidsoul: Yup saw that, I'll look into it later as soon as I can 😉

PR #260, opened 2017-09-29
Feature/259 urlrequest httpbody function

mplorentz:
### Checklist

- [x] I've checked that all new and existing tests pass
- [x] I've updated the documentation if necessary
- [x] I've added an entry in the CHANGELOG to credit myself

### Description

Fixed the URLRequest.ohhttpStubs_httpBody in Swift 3 and 4.

### Motivation and Context

Fixes #259 and allows users to make assertions on the HTTP POST body of a `URLRequest` in `OHHTTPStubsTestBlock`s and `OHHTTPStubsResponseBlock`s.

Sorry about the sloppy commits. I didn't want to download Xcode 7 again so I did a little guess and check with Travis.

PR #265, opened 2017-10-20
Add matcher "hasJsonBody"

pimnijman:
### Checklist

### Description

I've added a matcher called `hasJsonBody` that  tests whether the `NSURLRequest` body contains a JSON object with the same keys and values.

### Motivation and Context

Since it shouldn't matter in what way the attributes of a JSON object are ordered, trying to match the URL request's body using `hasBody` isn't very handy when it comes to verifying JSON content. That is why I've created a matcher that takes a JSON object and matches is with whatever JSON object it can find in the URL request's body.

I've added two methods in `SwiftHelpersTests.swift` that test a handful of scenarios. Please check them out and let me know if you think I overlooked some important scenarios

### DIFF
--- a/OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift
+++ b/OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift
@@ -401,7 +401,10 @@ public func hasHeaderNamed(_ name: String, value: String) -> OHHTTPStubsTestBloc
 #if swift(>=3.0)
 public func hasJsonBody(_ jsonObject: [AnyHashable : Any]) -> OHHTTPStubsTestBlock {
   return { req in
-    guard let jsonBody = (try? JSONSerialization.jsonObject(with: (req as NSURLRequest).ohhttpStubs_HTTPBody(), options: [])) as? [AnyHashable : Any] else {
+    guard
+      let httpBody = req.ohhttpStubs_httpBody,
+      let jsonBody = (try? JSONSerialization.jsonObject(with: httpBody, options: [])) as? [AnyHashable : Any]
+    else {
       return false
     }
     return NSDictionary(dictionary: jsonBody).isEqual(to: jsonObject)
--- a/OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift
+++ b/OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift
@@ -392,6 +392,23 @@ public func hasHeaderNamed(_ name: String, value: String) -> OHHTTPStubsTestBloc
   }
 #endif
 
+/**
+ * Matcher testing that the `NSURLRequest` body contains a JSON object with the same keys and values
+ * - Parameter jsonObject: the JSON object to expect
+ *
+ * - Returns: a matcher that returns true if the `NSURLRequest`'s body contains a JSON object with the same keys and values as the parameter value
+ */
+#if swift(>=3.0)
+public func hasJsonBody(_ jsonObject: [AnyHashable : Any]) -> OHHTTPStubsTestBlock {
+  return { req in
+    guard let jsonBody = (try? JSONSerialization.jsonObject(with: (req as NSURLRequest).ohhttpStubs_HTTPBody(), options: [])) as? [AnyHashable : Any] else {
+      return false
+    }
+    return NSDictionary(dictionary: jsonBody).isEqual(to: jsonObject)
+  }
+}
+#endif
+
 // MARK: Operators on OHHTTPStubsTestBlock
 
 /**
--- a/OHHTTPStubs/UnitTests/Test Suites/SwiftHelpersTests.swift
+++ b/OHHTTPStubs/UnitTests/Test Suites/SwiftHelpersTests.swift
@@ -454,6 +454,72 @@ class SwiftHelpersTests : XCTestCase {
 #endif
   }
 
+#if swift(>=3.0)
+  func testHasJsonBodyIsTrue() {
+    let jsonObjects = [
+      // Exact match
+      (["foo": "bar", "baz": 42, "qux": true],
+       ["foo": "bar", "baz": 42, "qux": true]),
+      // Changed attribute order
+      (["foo": "bar", "baz": 42, "qux": true],
+       ["qux": true, "foo": "bar", "baz": 42]),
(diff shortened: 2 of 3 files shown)


OHHTTPStubs/Sources/Swift/OHHTTPStubsSwift.swift
```diff
@@ -392,6 +392,23 @@ public func hasHeaderNamed(_ name: String, value: String) -> OHHTTPStubsTestBloc
...
+/**
+ * Matcher testing that the `NSURLRequest` body contains a JSON object with the same keys and values
+ * - Parameter jsonObject: the JSON object to expect
+ *
+ * - Returns: a matcher that returns true if the `NSURLRequest`'s body contains a JSON object with the same keys and values as the parameter value
+ */
+#if swift(>=3.0)
+public func hasJsonBody(_ jsonObject: [AnyHashable : Any]) -> OHHTTPStubsTestBlock {
+  return { req in
+    guard let jsonBody = (try? JSONSerialization.jsonObject(with: (req as NSURLRequest).ohhttpStubs_HTTPBody(), options: [])) as? [AnyHashable : Any] else {
```
AliSoftware: We very recently fix that need for the cast to `NSURLRequest` (see https://github.com/AliSoftware/OHHTTPStubs/pull/260) so `as` can be removed to simplify this expression :)

OHHTTPStubs/UnitTests/Test Suites/SwiftHelpersTests.swift
```diff
@@ -454,6 +454,72 @@ class SwiftHelpersTests : XCTestCase {
...
+      // Changed attribute order
+      (["foo": "bar", "baz": 42, "qux": true],
+       ["qux": true, "foo": "bar", "baz": 42]),
+      // Nested objects
+      (["foo": "bar", "baz": ["qux": true, "quux": ["spam", "ham", "eggs"]]],
+       ["foo": "bar", "baz": ["qux": true, "quux": ["spam", "ham", "eggs"]]]),
+      // Nested objects with changed attribute order
+      (["foo": "bar", "baz": ["qux": true, "quux": ["spam", "ham", "eggs"]]],
+       ["foo": "bar", "baz": ["quux": ["spam", "ham", "eggs"], "qux": true]]),
+    ]
```
AliSoftware: The problem with that approach is that dictionaries have no guaranteed order (and that the whole point of your PR). So there's no guarantee that Swift won't make all those dictionaries exactly the same internally, and then making the JSON serialized version of all those exactly the same (e.g. one might even imagine that `JSONSerialization.data(withJSONObject:)` could iterate over the dictionaries' keys in alphabetical order…)

To solve that, I think you'll instead need to use the already-converted-to-JSON String representations as a test set instead.

Which would have the additional benefit of proving that your matcher is also robust against different indentations of JSON bodies.

One way to do that is to use string literals of JSON representations to test
```
let jsonObjects = [
  "{ \"foo\": \"bar\", \n\"baz\": 42,      \"qux\": true }",
…
```
But for readability of the unit tests, maybe outsourcing those texts/JSON representation in fixture files would be nicer.

OHHTTPStubs/UnitTests/Test Suites/SwiftHelpersTests.swift
```diff
@@ -454,6 +454,72 @@ class SwiftHelpersTests : XCTestCase {
...
+      XCTAssertTrue(matchesJsonBody)
+    }
+  }
+#endif
+
+#if swift(>=3.0)
+  func testHasJsonBodyIsFalse() {
+    let jsonObjects = [
+      // Changed value
+      (["foo": "bar"],
```
AliSoftware: Same idea here of course

OHHTTPStubs/UnitTests/Test Suites/SwiftHelpersTests.swift
pimnijman: Your suggestion makes sense, but just to be clear; what you're suggesting is the following:

```
    let jsonStrings = [
      ("{ \"foo\": \"bar\", \"baz\": 42, \"qux\": true }",
       ["foo": "bar", "baz": 42, "qux": true])
      // Etc...
    ]
    
    for (jsonString, expectedJsonObject) in jsonStrings {
      var req = URLRequest(url: URL(string: "foo://bar")!)
      req.httpBody = jsonString.data(using: .utf8)
      let matchesJsonBody = hasJsonBody(expectedJsonObject)(req)
      
      XCTAssertTrue(matchesJsonBody)
    }
```

AliSoftware: Yup, exactly 👍

pimnijman:
@AliSoftware, should I create a new pull request or is there a way to resubmit this one? I'm new to this. :)

OHHTTPStubs/UnitTests/Test Suites/SwiftHelpersTests.swift
```diff
@@ -454,6 +454,69 @@ class SwiftHelpersTests : XCTestCase {
 #endif
   }
 
+#if swift(>=3.0)
+  func testHasJsonBodyIsTrue() {
+    let jsonStringsAndObjects = [
```
AliSoftware: Nitpicking, but just realized: why make that an array of tuples, instead of a Dictionary?

Wouldn't change a think in your `for` loop below, but would be more common Swift and would make it order-independent.

pimnijman: I'm not sure I agree. Of course I could convert this array to a dictionary but this would mean I can use each JSON sting only once. Unless I've misinterpreted your comment.

```
let dict = [ "{ \"foo\": \"bar\" }", ["foo": "bar" ] ]
```

AliSoftware: Oh I didn't realize you used some test input strings multiple times, good point

AliSoftware:
@pimnijman No need to create a new PR 😃

A pull request is a request to merge a branch into another branch, so if you push new commits to this branch like you just did, even after creating the initial PR, they'll be included in the PR as well, no need to create a new one 😃

I'll merge as soon as the CI (Travis) is ready

PR #267, opened 2017-11-20
fix xcode 9 warnings and recommended settings

NikSativa:
### Checklist


AliSoftware:
Builds seem to be failing due to `SWIFT_VERSION` not being set appropriately?

NikSativa:
did update both projects and all targets/framworks to the latest version of xcode 9.1 and swift version 4.0. all warnings and errors are fixed

AliSoftware:
Doesn't Xcode 9.1 require Swift 4.0.2 instead of 4.0?

NikSativa:
no, 4.x required
4.0.2 is the minor update

AliSoftware:
Yeah but for people using Carthage, which builds the framework based on the settings in the Xcode project before linking it to the integrating product, they'll probably want the framework to be built with the latest Swift ABI & version, so 4.0.2 ?

NikSativa:
see screenshot, xcode contains only 1 integrated version inside
thats why xcode 9.0 will contain 4.0, but xcode 9.1 - 4.0.2
and inside project/target setting the version will be the same 4.0
see screenshot:
screen shot 2017-11-20 at 20 13 36

AliSoftware:
:+1: 
(Food disclosure: I only checked from the github web interface as I was on my phone and not at a computer, sorry)

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1468,7 +1451,9 @@
 					"$(inherited)",
 				);
 				INFOPLIST_FILE = "UnitTests/UnitTests-Info.plist";
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
```
Liquidsoul: Making this change on the deployment target and the one below for macOS, drops the support for previous OS.
I am not sure that this is what we want because this is not what is exposed in the pod spec file (we support 7.0).
My point is that we could leave the Xcode 9 warning exposing the fact that it does not support iOS prior to 8.0 and macOS prior to 10.9 because, maybe, some people still use the previous versions.

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1685,18 +1686,22 @@
 				GCC_NO_COMMON_BLOCKS = YES;
 				INFOPLIST_FILE = "UnitTests/UnitTests-Info.plist";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				MACOSX_DEPLOYMENT_TARGET = 10.9;
 				MTL_ENABLE_DEBUG_INFO = NO;
 				PRODUCT_BUNDLE_IDENTIFIER = "alisoftware.OHHTTPStubs-tvOS-Fmk-Tests";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = appletvos;
-				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
```
Liquidsoul: Was this setting removed when supporting the Xcode recommended settings?

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1599,22 +1592,25 @@
...
 				WRAPPER_EXTENSION = xctest;
 			};
 			name = Release;
 		};
 		725CD9AE1A9EB65200F84C8B /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				CLANG_WARN_UNREACHABLE_CODE = YES;
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
```
Liquidsoul: Is this necessary to change the code signing settings in the project? Was this done by the automatic process of Xcode?

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
NikSativa: few lines below:
MACOSX_DEPLOYMENT_TARGET = 10.9;

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
NikSativa: all of this changes was done automatically by "update xcode project to recommended settings"

"project level" is more preferred instead of framework/target settings, because project can be changed by every developer instead of only code owner
example, i don't want to see any warnings in my project and than i can setup any MY project settings which will inherited by YOURs target. proposal: ios/macos version must be removed and xcode will use "default" version based on the parent project

and as proposal you must review your demo/project/target/framework settings, because as you see many of them are old legacy from the previous versions

jeffctown:
@NikSativa - Hey there!  We just released version 7 of OHHTTPStubs which is compatible with Xcode 10 and Swift 4.x, so I am closing this PR out, but I wanted to say thank you for submitting it! We really appreciate the community support!

PR #271, opened 2018-01-14
containsAtLeastQueryParams: A matcher that checks only the target params

crsantos:
### Checklist

### Description

This is my proposal for a matcher that checks *ONLY* the target parameters, and does not fail when more parameters are found. Just like `containsAtLeastQueryParams`, but it doesn't break when more parameters are found.

### Motivation and Context

My motivation was that I need to stub a request but I didn't want to handle all the parameters, just 2 or 3. 
So , this matches **at least** the target parameters, and ignores the others, if there are more than the target ones, ignores them, validating only the passed ones.

I don't know if I should bump spec version, if not, please say so, so I can revert.

Thanks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # OHHTTPStubs — CHANGELOG
 
+## [6.1.1](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/6.1.1)
+
+* Added `containsAtLeastQueryParams` matcher.  
+  [@crsantos](https://github.com/crsantos)
+  
 ## [6.1.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/6.1.0)
 
 * Updated deployment target for the pod to 7.0 to remove warning for old APIs.  
--- a/OHHTTPStubs.podspec
+++ b/OHHTTPStubs.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
 
   s.name         = "OHHTTPStubs"
-  s.version      = "6.1.0"
+  s.version      = "6.1.1"
 
   s.summary      = "Framework to stub your network requests like HTTP and help you write network unit tests with XCTest."
   s.description  = <<-DESC.gsub(/^ +\|/,'')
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # OHHTTPStubs — CHANGELOG
 
+## [6.1.1](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/6.1.1)
```
AliSoftware: Please just use a new `## Master` section for now instead of a future version number as we don't know when we'll release a new version and what its number will be yet.

(besides, your PR _adds_ a feature, while bumping the 3rd number of a version means fixing a bug and me feature requires bumping the minor version number not the patch —read semver.org to learn more if you like)

OHHTTPStubs.podspec
```diff
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
 
   s.name         = "OHHTTPStubs"
-  s.version      = "6.1.0"
+  s.version      = "6.1.1"
```
AliSoftware: You can revert this change, my scripts will take care of updating the podspec with the future version number when we decide to release a new version and determine what that version number should be 😉

OHHTTPStubs/UnitTests/Test Suites/SwiftHelpersTests.swift
```diff
@@ -376,6 +376,39 @@ class SwiftHelpersTests : XCTestCase {
...
+      "foo://bar;lang=en&q=test": false,
+
+      "foo://bar?q=test&lang=en": true, // matches at least all params
+      "foo://bar?lang=en&q=test": true, // matches at least all params in any order
+      "foo://bar?q=en&lang=test": false, // param keys exist but values mismatch
+      "foo://bar?q=test&lang=en&&wizz=fuzz": true,
+      "foo://bar?wizz=fuzz&lang=en&&q=test": true,
+      "?q=test&lang=en": true,
+      "?lang=en&flag&empty=&q=test": true,
+    ]
```
AliSoftware: 👍 Good test set

CHANGELOG.md
crsantos: Thanks, will update

OHHTTPStubs.podspec
crsantos: 👌

crsantos:
allowMoreParams as a default param looks way better!
I’ll update. 
Actually PR is closed but I’ll reopen soon when aí finish all updates. 
Thanks for reviewing

PR #279, opened 2018-05-18
Update for Xcode 9.3

rckoenes:
### Checklist

- [✓] I've checked that all new and existing tests pass
- [✓] I've updated the documentation if necessary
- [✓] I've added an entry in the CHANGELOG to credit myself

### Description

Update the project so it can be used with Xcode 9.3 and swift 4
Also fixed some static analyser warning.

### Motivation and Context
We have many framework which where on swift 4 and carthage did not want to build this 3.1 framework.


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # OHHTTPStubs — CHANGELOG
 
+## [6.1.1](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/6.1.1)
+* Update the project setting of swift to 4.0
+  [@rckoenes](https://github.com/rckoenes/OHHTTPStubs)
+* Fixed some `nullable` warning form the static analyzer
+  [@rckoenes](https://github.com/rckoenes/OHHTTPStubs)
+* Fixed a warning related to a property have copy while use a mutable array, whicih will turn it into an immutable array
```
Liquidsoul: Small typo here `whicih`.
In fact, I would rephrase that a bit : "Fixed a static analyser warning related to a property having the `copy` trait while using a mutable array, ..."

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1329,7 +1331,7 @@
 				MODULEMAP_FILE = "$(SRCROOT)/Supporting Files/module.modulemap";
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
-				SWIFT_VERSION = 3.0;
+				SWIFT_VERSION = 4.0;
```
Liquidsoul: I think that we must not bump the swift version here because we are still supporting Swift 3.x and the CI still run tests using Xcode 8.3.

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1387,6 +1391,7 @@
 		09110A6519805F4800D175E4 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				CURRENT_PROJECT_VERSION = 6.1.1;
```
Liquidsoul: If you want to integrate a version bump of the library, you should do it at the Project level, not for each Target.
If you use the "Levels" view in the Build Settings you'll see that the Project's version is still set to `6.1.0`.
We have the version at the Project level to share this information and change it only in one place.
To fix this, change the version in the `OHHTTPStubs` project and remove the override on each target.

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
rckoenes: The hole problem is that I can not use the Swift 3 version in carthage because Xcode fails if I include a framework build with Swift 3

Liquidsoul: I understand your problem but this is something we cannot do without considering the fact that we still support Swift 3 and that the CI needs to be able to check that for us.
AFAIK, the support of multiple Swift versions is not possible with Carthage, we've even opened an [issue](https://github.com/Carthage/Carthage/issues/1445) about this that was not resolved.
See [this section](https://github.com/AliSoftware/OHHTTPStubs#using-the-right-swift-version-for-your-project) of the README for our proposed workarounds for Carthage.

ikesyo: > AFAIK, the support of multiple Swift versions is not possible with Carthage, we've even opened an issue about this that was not resolved.

Xcode 9 and above support Swift 3.x and Swift 4 for each-target basis, so this should not be an issue nowadays.

rckoenes: I'm fully aware that support Swift 3 is support by Xcode and would make this project work on the most version of Xcode. 
We just have an automated build system that required some modules in Swift 4 and it did not want to build with a Swift 3 module in place.

jeffctown:
@rckoenes - Hey there!  We just released version 7 of OHHTTPStubs which is compatible with Xcode 10 and Swift 4.x, so I am closing this PR out, but I wanted to say thank you for submitting it! We really appreciate the community support!

PR #286, opened 2018-08-02
Add support to WatchOS

linksmt:
### Checklist

- [ ] I've checked that all new and existing tests pass
- [ ] I've updated the documentation if necessary

### DIFF
--- a/OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
+++ b/OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
@@ -1768,7 +1768,7 @@
 				PRODUCT_NAME = OHHTTPStubs;
 				SDKROOT = watchos;
 				SKIP_INSTALL = YES;
-				TARGETED_DEVICE_FAMILY = "1,2";
+				TARGETED_DEVICE_FAMILY = 4;
 			};
 			name = Debug;
 		};
@@ -1790,7 +1790,7 @@
 				SDKROOT = watchos;
 				SKIP_INSTALL = YES;
 				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
-				TARGETED_DEVICE_FAMILY = "1,2";
+				TARGETED_DEVICE_FAMILY = 4;
 			};
 			name = Release;
 		};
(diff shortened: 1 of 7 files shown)


OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -374,6 +403,7 @@
 				09110A4319805F4800D175E4 /* Frameworks */,
 				09110A4219805F4800D175E4 /* Products */,
 				71E7CDE1C6A8345F6C70E7D1 /* Pods */,
+				C8EDD4532111DF8200E60E4E /* OHHTTPStubs iOS Framework copy-Info.plist */,
```
AliSoftware: Please rename that file appropriately for consistency :wink:

tfonfara:
I'm a bit concerned if those changes are enough, as there is on several places code like `#if defined(__IPHONE_7_0)`, that should be extended with `__WATCHOS_3_0` or similar

AliSoftware:
good point.

AliSoftware:
Ping @linksmt about the `#if defined(__IPHONE_7_0)` remark above 😉

tfonfara:
I tried to come up with a solution by myself, but things are petty weird. I've added all necessary precompiler defines, but OHHTTPStubsProtocol isn't working properly.
`canInitWithRequest` is being called and also returns true, but neither `initWithRequest` nor `startLoading` gets called. I also tried with a simple approach in a fresh project to implement NSURLProtocol but found same behaviour.

linksmt:
Can you accept merge? 
Thanks

jeffctown:
@linksmt thank you for submitting this!  

Everything looks great for the added target, but I think we should probably have some tests to go along with it so we can make sure we don't break it.  😄 

Would you be comfortable adding some tests to the watchOS target?  

I think it would also be helpful if you update the `CHANGELOG.md` to give yourself credit and update the original description of this PR.

1ec5:
> Would you be comfortable adding some tests to the watchOS target?

Unless I’m missing something, Xcode doesn’t support watchOS test targets yet.

Juanpe:
Hi! @AliSoftware, what is the status of this PR? I mean, `OHHTTPStubs` is still not supporting **watchOS**, doesn't it?

Issue #230, opened 2016-11-30
Redirect stubbing not working reliably


nevil:
First of all, thanks for this great tool!
Sorry for the disorganized report

I've been having some issue that our redirect unit tests occasionally fail.
In the latest release I saw a fix that I was hoping to solve my problem ( #224) but unfortunately not.

Now I have a way to make the redirect stubbing fail reliably. A small modification to OHHTTPStubs.m:
```
[client URLProtocol:self wasRedirectedToRequest:redirectRequest redirectResponse:urlResponse];
+++ usleep(1000000)
```

With the above we get the failure 100% of the time.
I added a test to NSURLSessionTests.m:
```
- (void)test_Redirect
{
    if ([NSURLSession class])
    {
        NSURLSession *session = [NSURLSession sharedSession];
        
        NSDictionary* json = @{@"Success": @"Yes"};
        [self _test_redirect_NSURLSession:session jsonForStub:json completion:^(NSError *errorResponse, NSHTTPURLResponse *redirectResponse, id jsonResponse) {
            XCTAssertNil(errorResponse, @"Unexpected error");
            XCTAssertNil(redirectResponse, @"Unexpected redirect response received");
            XCTAssertEqualObjects(jsonResponse, json, @"Unexpected data received");
        }];
    }
    else
    {
        NSLog(@"/!\\ Test skipped because the NSURLSession class is not available on this OS version. Run the tests a target with a more recent OS.\n");
    }
}
```

After inserting the sleep the data task completion handler starts  receiving the redirect response instead of the response after the redirect.

I can fix it by not calling didReceiveResponse for the redirect (301 code) but based on #175 this does not seem to be a valid solution

I will investigate further but wanted to create a ticket in case anyone already have a clue about the root cause.

AliSoftware:
🤔 Interesting, thanks for the report.
No guarantee I'll have much time to look at it soon, but will keep an eye on it and try to check it some time 👍

AliSoftware:
Hey @nevil did you manage to investigate this issue further and understand what was causing it?

nevil:
@AliSoftware 
Sorry. I didn't have the time yet to find the real cause.
I have temporarily changed the requestTime to 0.2 and responseTime to 0.5 (arbitrarily selected) and it prevents the issue from happening for me.

This does slow down my unit test execution a bit (though not more than 10 seconds :-) ) so I hope to have time to investigate soon.

AliSoftware:
[...]

PR #289, opened 2018-09-15
Drop special headers from redirected requests

sberrevoets:
### Checklist

### Description

The change in https://github.com/AliSoftware/OHHTTPStubs/pull/263 means a redirected request had the same headers as the original request (because it was a copy and not a brand new `NSURLRequest` instance).

While the new behavior was correct, it didn't account for "special headers" that URLSession "handles". The docs aren't entirely clear what that means, but we've at least confirmed it removes the "Authorization" headers as that poses a security risk, so this PR removes all those headers + a test.

### DIFF
--- a/OHHTTPStubs/Sources/OHHTTPStubs.m
+++ b/OHHTTPStubs/Sources/OHHTTPStubs.m
@@ -368,6 +368,24 @@ static NSTimeInterval const kSlotTime = 0.25; // Must be >0. We will send a chun
     return nil;
 }
 
+/** Drop certain headers in accordance with
+ * https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1411532-httpadditionalheaders
+ */
+- (NSMutableURLRequest *)clearAuthHeadersForRequest:(NSMutableURLRequest *)request {
+    NSArray* authHeadersToRemove = @[
+                                     @"Authorization",
+                                     @"Connection",
+                                     @"Host",
+                                     @"Proxy-Authenticate",
+                                     @"Proxy-Authorization",
+                                     @"WWW-Authenticate"
+                                     ];
+    for (NSString* header in authHeadersToRemove) {
+        [request setValue:nil forHTTPHeaderField:header];
+    }
+    return request;
+}
+
 - (void)startLoading
 {
     self.clientRunLoop = CFRunLoopGetCurrent();
@@ -446,17 +464,10 @@ static NSTimeInterval const kSlotTime = 0.25; // Must be >0. We will send a chun
                             //Preserve the original request method and body, and set the new location URL
                             mReq = [self.request mutableCopy];
                             [mReq setURL:redirectLocationURL];
-
-                            // Drop certain headers in accordance with
-                            // https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1411532-httpadditionalheaders
-                            [mReq setValue:nil forHTTPHeaderField:@"Authorization"];
-                            [mReq setValue:nil forHTTPHeaderField:@"Connection"];
-                            [mReq setValue:nil forHTTPHeaderField:@"Host"];
-                            [mReq setValue:nil forHTTPHeaderField:@"Proxy-Authenticate"];
-                            [mReq setValue:nil forHTTPHeaderField:@"Proxy-Authorization"];
-                            [mReq setValue:nil forHTTPHeaderField:@"WWW-Authenticate"];
+                            
+                            mReq = [self clearAuthHeadersForRequest:mReq];
+                            
                             redirectRequest = (NSURLRequest*)[mReq copy];
-
                             break;
                         }
                         default:
--- a/OHHTTPStubs/Sources/OHHTTPStubs.m
+++ b/OHHTTPStubs/Sources/OHHTTPStubs.m
@@ -360,12 +360,12 @@ static NSTimeInterval const kSlotTime = 0.25; // Must be >0. We will send a chun
 
 + (NSURLRequest *)canonicalRequestForRequest:(NSURLRequest *)request
 {
-	return request;
+    return request;
 }
 
 - (NSCachedURLResponse *)cachedResponse
 {
-	return nil;
+    return nil;
 }
 
 - (void)startLoading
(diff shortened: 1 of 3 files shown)


keith:
I think these test failures are related to your changes @sberrevoets ? It looks like master is mostly green (besides some flakes)

sberrevoets:
I was able to reproduce this once (via `rake`, I never saw them fail in Xcode 10), but now even `rake` is succeeding locally.

@AliSoftware any idea what might be going on?

AliSoftware:
I think the build failures might be related to the race condition bug we discovered in URL session in https://github.com/AliSoftware/OHHTTPStubs/issues/230#issuecomment-380931933 (don't hesitate to duplicate the radar to help us make Apple fix it btw)

sberrevoets:
Ah thanks for the heads up, I've duped the radar. What would you like to see here in the meantime?

sberrevoets:
Friendly ping @AliSoftware :)

AliSoftware:
Sorry for the late reply.
I honestly don't really know. I mean those redirect tests are probably always gonna fail until that Apple bug gets fixed or we find a workaround.
So maybe we should merge that anyway given that the failures are not our fault?

Or a better solution might be to introduce a small delay in the redirection (like what we done in the other PR to confirm the bug was a race condition) as it seemed to at least reduce a lot the bug from happening; even if that means introducing a delay and slowing down stuff a little, that's probably better than failing 70% of the time…?

sberrevoets:
Yeah that's a very good question. For me it'd probably depend on how much it fails. If it really is 70% of the time right now then adding a delay seems worth it. If it's every now and then, it might be possible to (somehow) auto-retry the affected tests, or take a simpler approach of adding a comment explaining the problem.

I'm also open to alternatives, framework bugs like these are a huge PITA 😕

jeffctown:
@sberrevoets - we're reviewing all outstanding PRs before making a release, and then updating the version of Swift being used.  the hope is to do all of this before Xcode 10.2 comes out, so our timeline is a bit tight.  

If you are still interested in getting this in, please resolve the conflicts. If you are not interested, having that feedback would be helpful too.

sberrevoets:
Happy to resolve conflicts (though can't promise I'll have time for that before the 10.2 release), but _how_ depends on what approach is preferable. Based on the conversation above I'm not sure if there's a clear path forward.

jeffctown:
@sberrevoets - the flaky tests were disabled in CI (#301 & #302), so you don't need to worry about that part.  We only need the conflicts resolved.

sberrevoets:
Ah I missed that, that makes resolving conflicts a bit easier. Pushed an update but also gave edit access just in case.

OHHTTPStubs/Sources/OHHTTPStubs.m
```diff
@@ -442,13 +442,23 @@ - (void)startLoading
...
                             [mReq setURL:redirectLocationURL];
+
+                            // Drop certain headers in accordance with
+                            // https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1411532-httpadditionalheaders
+                            [mReq setValue:nil forHTTPHeaderField:@"Authorization"];
+                            [mReq setValue:nil forHTTPHeaderField:@"Connection"];
+                            [mReq setValue:nil forHTTPHeaderField:@"Host"];
+                            [mReq setValue:nil forHTTPHeaderField:@"Proxy-Authenticate"];
+                            [mReq setValue:nil forHTTPHeaderField:@"Proxy-Authorization"];
+                            [mReq setValue:nil forHTTPHeaderField:@"WWW-Authenticate"];
```
AliSoftware: This could maybe be a `for header in @[@"…",@"…",…]` loop instead?

jeffctown: I pushed a commit to put this into it's own method since where it is now is already a bit large.

jeffctown:
@AliSoftware I updated this branch with the for loop, and I still see the conditional tests.

AliSoftware:
I was talking about a new test case that @sberrevoets wrote in his initial commits, and that I remember about from when I reviewed the diff on that PR before the conflict resolution today. That new test case which was added in that PR was essentially creating a request, with all the headers of interest here + another header — named `"preserved"` iirc —, and after using that dummy request and redirecting it, did `XCTAssertNil()` on all proper headers to check they were properly removed and an `XCTAssertEqual()` on the one named `"preserved"` to check it wasn't removed.

That test case seems not to be there anymore (or is my GitHub view of the PR diff on my phone making me missing something?). It probably was removed by the force-push done by @sberrevoets today I guess. While I think it would be worth to have that test case back (even if it ends up not being run on CI because we disabled them there, but they would still be able to be run locally in Xcode)

jeffctown:
Oh! My bad man. I thought you meant the buggy test. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

sberrevoets:
I did remove that test case because it was one of the conflicts. Working on adding that back now.

sberrevoets:
Added that back and ran it locally - it still passed.

OHHTTPStubs/UnitTests/Test Suites/NSURLSessionTests.m
```diff
@@ -360,6 +360,51 @@ - (void)test_NSURLSessionDefaultConfig_MethodAndDataRetentionOnRedirect
...
+                NSURLSession *session = [NSURLSession sessionWithConfiguration:config delegate:delegate delegateQueue:nil];
+
+                NSDictionary *headers = @{
+                                          @"Authorization": @"authorization",
+                                          @"Connection": @"connection",
+                                          @"Host": @"host",
+                                          @"Proxy-Authenticate": @"proxy-authenticate",
+                                          @"Proxy-Authorization": @"proxy-authorization",
+                                          @"WWW-Authenticate": @"www-authenticate",
+                                          @"Preserved": @"preserved"
```
AliSoftware: Just because say we change the implementation one day and accidentally make the new implementation behave ( wrongly) as "remove all headers but the last" or "only keep one header" (instead of doing what it's supposed to do like today), I would try to put one or two more "headers to be preserved" in there _and_ put them in the middle of that list instead of all in the end 😉 #paranoiaModeOn

sberrevoets: But since it's a dictionary order isn't guaranteed anyway? Is that different under the NSURLSession hood?

AliSoftware: Oh good point!

Though I suspect that dictionary _literals_, even though they don't guarantee order by contract as it's not a expected requirement for dicts, might end up always being iterated in the same deterministic order in practice (because of current implantation details of dictionary literals in ObjC)?

sberrevoets: Yeah it doesn't hurt to add - done. (In a separate commit for reviewing purposes, happy to rebase at the end if you'd like.)

AliSoftware:
Thanks @sberrevoets :tada:

sberrevoets:
Thanks for the reviews and suggestions!

PR #292, opened 2018-10-27
update swift demo compatibility with swift 4.2

hellensoloviy:
### Checklist

- [x] I've added an entry in the CHANGELOG to credit myself (I'm not sure if I add it in correct place but I add, fix it if needed please)

### Description

Update swift example project to swift 4.2 version and fix compatibility issue; 

### Motivation and Context

It helps to build and run project on last swift an xcode 

no, there is no open issue 

build and run the example app. Run all tests also


README.md
```diff
@@ -73,7 +73,7 @@ _(There are also other ways to perform a similar task, including using `curl -is
 
 * `OHHTTPStubs` is compatible with **iOS5+**, **OS X 10.7+**, **tvOS**.
 * `OHHTTPStubs` also works with `NSURLSession` as well as any network library wrapping them.
-* `OHHTTPStubs` is **fully compatible with Swift 2.2, 2.3, 3.0 and 3.1**.
+* `OHHTTPStubs` is **fully compatible with Swift 2.2, 2.3, 3.0, 3.1 and 4.2**.
```
AliSoftware: Maybe at this point we'd just state "from 2.2 to 4.2"… although I'm not sure that's really true. We might have dropped 2.x support at some point and forgot about updating that part of the README maybe? Would be good to check if we're still compatible with 2.x… or remove that from the sentence altogether as probably nobody uses 2.2 anymore…

Examples/Swift/AppDelegate.swift
```diff
@@ -13,7 +13,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
 
-    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
```
AliSoftware: It would be nice to add a compatibility alias so that the code would still compile with older Xcode

Something like:
```
#if !swift(>=4.2)
extension UIApplication {
  static let LaunchOptionKey = UIApplicationLaunchOptionKey
}
#endif
```

hellensoloviy:
Hello! 
Thanks for the answer on my PR :)
I have made some changes based on your comments, so u can look at it if u wanna) 

Have a nice evening!

README.md
```diff
@@ -158,10 +158,11 @@ For more info, see [CocoaPods/CocoaPods#5540](https://github.com/CocoaPods/Cocoa
...
 
-The project is set up with `SWIFT_VERSION=3.0` on `master`.
+The project is set up with `SWIFT_VERSION=4.2` on `master`.
 
 This means that the framework on `master` will build using:
 
+* Swift 4.0/4.2 on Xcode 10
 * Swift 3.1 on Xcode 8.3
 * Swift 3.0 on Xcode 8.2
 * Swift 2.2/2.3 on Xcode 7.x.
```
AliSoftware: I wonder if that's still true?

iirc, that worked because Xcode 7 didn't know about (thus ignored) the `SWIFT_VERSION=3.0` build setting and thus used 2.2, while Xcode 8 took it into account and used it.

But now that Xcode 8, 9 and 10 all know about `SWIFT_VERSION` and use it, I wonder what happens if we try to build a project with `SWIFT_VERSION=4.2` using an earlier Xcode which doesn't know that version of Swift? Does it error? Does it fallback to the latest version of Swift it knows?

As a consequence, wouldn't that change risk that OHHTTPStubs wouldn't be compatible with earlier Xcode versions especially if integrated with Carthage, which uses the Xcode project to build the framework?

Any Carthage user have an idea about that? @ikesyo maybe?

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1735,6 +1746,7 @@
 				SDKROOT = appletvos;
 				SKIP_INSTALL = YES;
 				SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";
+				SWIFT_VERSION = 4.2;
```
AliSoftware: That means that the library itself is compatible with Swift 4.2 👍 but as a result it could be worth updating the `OHHTTPStubs.podspec` to add that `s.swift_version` to the list 😉

AliSoftware: Mmmh I see we actually never added that `s.swift_version` line to the podspec actually… might be the occasion to do so then

http://blog.cocoapods.org/CocoaPods-1.4.0/#swift-version-dsl

hellensoloviy:
Yes, it's better to add it! Thanks for noticing. I've added it)

README.md
codenameDuchess: I think there are a far larger percentage of developers on Xcode 10 using Swift 4.2 than there are on older version of Xcode like 7, or 8.

jeffctown:
@hellensoloviy - Hey there!  We just released version 7 of OHHTTPStubs which is compatible with Xcode 10 and Swift 4.x, so I am closing this PR out, but I wanted to say thank you for submitting it! We really appreciate the community support!

hellensoloviy:
Hi, @jeffctown 
Thanks for your comment & mention.
I'm glad my PR was of help :)

PR #298, opened 2019-03-06
Disabling troublesome unit test.

### LINKED ISSUE #230: Redirect stubbing not working reliably

jeffctown:
* Disabling troublesome unit test.
* Updating Readme with known issue.

### Checklist

### Motivation and Context

Updates according to #230 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## [Future release]
 * Enabled application extension API only.
   [@lightsprint09](https://github.com/lightsprint09)
+* Disabled a flaky redirect test and updated README with issue.
+  [@jeffctown](https://github.com/jeffctown)
+
 
 ## [6.1.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/6.1.0)
 
--- a/OHHTTPStubs/UnitTests/Test Suites/NSURLSessionTests.m
+++ b/OHHTTPStubs/UnitTests/Test Suites/NSURLSessionTests.m
@@ -290,44 +290,44 @@
     }
 }
 
-/**
+/** https://github.com/AliSoftware/OHHTTPStubs/issues/230
  Verify that redirects of different methods and status codes are handled properly and
  that we retain the HTTP Method for specific HTTP status codes as well as the data payload.
  **/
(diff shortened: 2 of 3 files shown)


AliSoftware:
Just dequeuing my notifications rn, I see you have decided to not merge this but just FYI if we were to fail to find a solution and decided to indeed disable the test for CI to be more reliable, I'd at least do it the same way I did for timing tests: wrap them in a big `#if` condition which would default to true, and make it false on CI by injecting the build setting during rake invocation on CI (see `DISABLE_TIMING_TESTS` or whatever I called that `#if`)

PR #299, opened 2019-03-06
Updating Travis to only build pushes on master to prevent double buil…

jeffctown:
### Checklist

- [ ] I've added an entry in the CHANGELOG to credit myself - this is a CI change so I didn't think it warranted a CHANGELOG entry.

### Description

Updating Travis configuration so that we don't build PR commits twice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: objective-c
 
+branches:
+  only:
+    - master
+
 before_install:
   - gem install xcpretty --no-rdoc --no-ri --no-document --quiet
 


AliSoftware:
Does that mean that when we do a PR from feature/x to master, the CI won't trigger until the PR is actually merged into master?

Doesn't that defeat the purpose of the CI, if we only know _after the fact_ (after merge) if it passes or not?

Or does this config means the opposite, like only build PRs targeting master but not PRs targeting other branches nor commits on master?

jeffctown:
@AliSoftware sorry for not being more clear with the initial comment.

Before this change, Travis was building every pull request and any commits added to an existing pull request twice.  One build for the `pr` and one for the `push`.  This was my original pull request for swift 4.2, but I saw the same behavior for every new pull request and every commit pushed to an existing pull request:

screen shot 2019-03-06 at 4 13 04 pm

When you click those links you see the same exact build is run twice.

This change **will not prevent** pull requests from being built before a merge, it simply means they will only be built once instead of twice.   We should still expect Travis to kick off a build for any new pull requests, any commits pushed to an existing pull request, and for any merges into master (for releases and to update the build status badge).  

This is a screenshot of the checks for this pull request, which you can see only builds once instead of twice:

screen shot 2019-03-06 at 4 19 57 pm

Notice how the `push` build is gone.  The reason a second build with the `push` label doesn't spin up is because of this change. 

If the duplicate builds were there for a reason, or if you don't like this change - I'm happy to revert it.  I just thought getting these next releases out would go a little quicker if we didn't have them.

AliSoftware:
Oh great, perfect explanation, makes total sense then 👍 thanks!

(It has been years since I played with Travis, as I moved to circle CI for all my other OSS repos 😅)

Issue #280, opened 2018-05-20
Unreliable redirect test


Liquidsoul:
### New Issue Checklist

- [x] I have tried with the latest version of OHHTTPStubs
- [x] I have read the [README](https://github.com/AliSoftware/OHHTTPStubs/blob/master/README.md)
- [x] I have read the [Using the right Swift Version of `OHHTTPStubs` for your project](https://github.com/AliSoftware/OHHTTPStubs#using-the-right-swift-version-of-ohhttpstubs-for-your-project) section
- [x] I have searched in the [existing issues](https://github.com/AliSoftware/OHHTTPStubs/issues?utf8=✓&q=is%3Aissue)
- [x] I have read [the OHHTTPStubs wiki](https://github.com/AliSoftware/OHHTTPStubs/wiki) to see if there wasn't a detailed page talking about my issue

### Environment

- version of OHHTTPStubs: `master`

### Issue Description

Following the introduction of #263 the test `- [NSURLSessionTests test_NSURLSessionDefaultConfig_MethodAndDataRetentionOnRedirect]` fails _most of the time_ (see @AliSoftware [comment](https://github.com/AliSoftware/OHHTTPStubs/pull/263#issuecomment-344105021)).
We need to investigate this issue so that we can have a reliable CI.

Liquidsoul:
Ok, after some investigations this seems related to https://github.com/AliSoftware/OHHTTPStubs/issues/230#issuecomment-378385218

I have experienced the same issues (redirect data without response object) for the failures.
So this seems like a `NSURLProtocol` bug.

So now I see several options:
- we find a workaround in the library to make this reliable (that seems unlikely to me)
- we find a workaround for the tests, but this would mean that client code can fail, and this occurs a lot in the tests in my experience
- we disable the tests and document that redirections are not correctly supported by `OHHTTPStubs` because of an Apple bug

AliSoftware:
😢

I think we'll have to go with solution 3 then. I think we could still keep the code handling the redirections, but just document properly that there are issues due to the Apple bug.

PR #301, opened 2019-03-06
Disabling troublesome unit test.

jeffctown:
Updating Readme with known issue.
Changelog.

### Checklist

### Description

Disabling flaky tests and updating README with the known issue.

### Motivation and Context

There has been an issue with redirects for several years that seems to be caused by an Apple bug.  This issue has been investigated multiple times, and the path that we want to take forward is to disable the buggy test and update the README with the known issue. 

Discussion on this topic has been mostly happening in #230 and #280 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## [Future release]
 * Enabled application extension API only.
   [@lightsprint09](https://github.com/lightsprint09)
+* Disabled a flaky redirect test and adding the known issue with redirects to the README.
+  [@jeffctown](https://github.com/jeffctown)
+  [#301](https://github.com/AliSoftware/OHHTTPStubs/pull/301)
+
 
 ## [6.1.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/6.1.0)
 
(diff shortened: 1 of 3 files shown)


AliSoftware:
As commented in #299 maybe we should have kept that test enabled by default when run on Xcode and only disable it on CI (the same way I did with timing tests)?

jeffctown:
@AliSoftware I think this is a great idea.  I can make put up a pull request with that change soon, and it makes way more sense than just disabling it globally.  

Would you prefer for me to wait for your approval on any of these pull requests before merging?  You're bringing up totally valid concerns on them.

AliSoftware:
In a dream world 🦄 I'd say if you could let me the day after you opened a PR to have a chance to review it before merging, that could be interesting (at least for the first PRs you take responsibility for) so that I can validate the direction… 
But given how busy I am with induction into my new job I can't guarantee that I'll be as fast to react or review all your other PRs (tonight was just a light evening, not expecting many ^^) 😓

So given how you seem to have good motivation and you're helping unlocking a situation that I did let stale for too long (And thanks again for that 🙏), sure if there's a PR that can wait a few hours or the day then maybe you could wait so I might take a look before you merge… but in any case if you see I'm not reacting after one day (while @liquidsoul approves the PR on his end), or a change is trivial, feel free to make things go forward without waiting for me — so that we don't continue having that stale situation on OHHTTPStubs that I kept postponing for too long for lack of  my own availability. (We could always revert a PR if needs be in case of emergency after all)

jeffctown:
@AliSoftware - Sounds reasonable to me.  👍 I can continue in this fashion.

PR #302, opened 2019-03-07
Conditional Compilation Fixes.

jeffctown:
Uncommenting buggy redirect tests.
Adding xcconfig to put project settings into.
Adding preprocessor macros to let unit test arguments through.
Adding default value of preprocessor macros to default them to run conditional tests when using Xcode.
Adding comments for xcconfig.
Updating Rake arguments.

### Checklist

### Description

My original goal was to add a previously commented unit test back to the test suite, and to enable it when testing in Xcode but disable it when testing in CI.  During this investigation, I also noticed that the Timing Tests were not working in this way like they were originally intended to.  I ended up fixing those tests in this PR too.

#### Timing Tests -

If you look at previous test runs, you see that timing tests were being run in Travis:

screen shot 2019-03-07 at 7 03 59 am

Even though the right Xcode argument was being passed:

screen shot 2019-03-07 at 7 04 15 am

This seems like a bug with the project setup.

I found that we needed to add a preprocessor macro to the project's build settings to let the `xcodebuild` argument through. 

I always like using `xcconfig` files to manage build settings, so I created a project level `xcconfig` with the build settings and commented the config.  Build settings that are defined in `xcconfig` files can always be overridden by passing in `xcodebuild` arguments.

#### Redirect Tests -

I followed the same pattern with the uncommented redirect tests.

#### End Result -

You can see here that both sets of tests are run when using Xcode:

screen shot 2019-03-07 at 9 11 00 am

And that neither are ran during CI:

screen shot 2019-03-07 at 9 22 17 am


OHHTTPStubs/OHHTTPStubsProject.xcconfig
```diff
@@ -0,0 +1,30 @@
...
+OHHTTPSTUBS_SKIP_TIMING_TESTS=0
+
+//   xcodebuild Example:
+//     xcodebuild -workspace OHHTTPStubs/OHHTTPStubs.xcworkspace -scheme "OHHTTPStubs iOS StaticLib" -sdk iphonesimulator -configuration Debug ONLY_ACTIVE_ARCH=NO OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1 -destination 'name=iPhone 7,OS=latest' clean build test
+//
+//   rake Example:
+//     rake ios['iOS StaticLib','latest','build-for-testing test-without-building',"OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1"]
+OHHTTPSTUBS_SKIP_REDIRECT_TESTS=0
+
+GCC_PREPROCESSOR_DEFINITIONS=OHHTTPSTUBS_SKIP_TIMING_TESTS=$(OHHTTPSTUBS_SKIP_TIMING_TESTS) OHHTTPSTUBS_SKIP_REDIRECT_TESTS=$(OHHTTPSTUBS_SKIP_REDIRECT_TESTS)
```
AliSoftware: Not sure that really matters in practice in our case but shouldn't we used `$(inherited)` in that definition? (I think the default in Xcode is `GCC_PREPROCESSOR_DEFINITIONS=DEBUG=1` in debug — while it's empty in release — so not using `$(inherited)` here means that debug builds don't have `DEBUG=1` set anymore? Not that we currently use `#if DEBUG` in OHHTTPStubs as far as I know but just for configuration consistency…

OHHTTPStubs/UnitTests/Test Suites/NSURLSessionTests.m
```diff
@@ -294,65 +294,69 @@ - (void)test_NSURLSessionDefaultConfig_notFollowingRedirects
...
+            [session finishTasksAndInvalidate];
+        }
+    }
+    else
+    {
+        NSLog(@"/!\\ Test skipped because the NSURLSession class is not available on this OS version. Run the tests a target with a more recent OS.\n");
+    }
+}
+#endif
 - (void)test_NSURLSessionEphemeralConfig
```
AliSoftware: (Just because I like nitpicking: this feels a bit packed, so if you have the occasion of a new commit in this PR, adding an extra empty line to separate the endif and the next function while you're at it might help us breathe 😅)

OHHTTPStubs/OHHTTPStubsProject.xcconfig
jeffctown: I'm honestly not sure how, but the config file looks like it's value resolves to have `DEBUG=1`, even though it's not in the file.

screen shot 2019-03-07 at 7 39 47 pm

I thought build settings always inherited from right to left in Xcode's levels view (^^ that view).. so I wouldn't expect the config file to inherit a setting from the project, but it does. 🤷‍♂️ 

Now that I think more about it - I normally don't mix project build settings and also have an xcconfig with them.  My opinion is normally that if you have an xcconfig you should put the entire value of that setting into the xcconfig instead of the project's build settings. 

@AliSoftware - What do you think of moving `DEBUG=1` to the xcconfig instead, so there is no project build setting for this one?

AliSoftware: Strange that it magically inherit project settings indeed 🧐
Moving DEBUG=1 in the xcconfig file makes sense, but would require two separate xcconfig files then (one for each config), just wondering if that's worth it 🤔

jeffctown: Yes, probably overkill.  So do you think we should add `$(inherit)` or leave it as is?

AliSoftware: Let's try `$(inherited)` at least for consistency, and validate it still works as expected (eg doesn't double-define `DEBUG=1` in the resolved build settings after all)

jeffctown: There were actually a few places where we had `DEBUG=1` twice, so I removed the target level preprocessor macros where needed.

OHHTTPStubs/UnitTests/Test Suites/NSURLSessionTests.m
jeffctown: Addressed.

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1463,10 +1467,6 @@
 				);
 				GCC_PRECOMPILE_PREFIX_HEADER = YES;
 				GCC_PREFIX_HEADER = "UnitTests/UnitTests-Prefix.pch";
-				GCC_PREPROCESSOR_DEFINITIONS = (
-					"DEBUG=1",
-					"$(inherited)",
-				);
```
jeffctown: These were added at the target level, and caused `DEBUG=1` to end up in the resolved macros twice.

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1519,10 +1519,6 @@
 				);
 				FRAMEWORK_VERSION = A;
 				GCC_ENABLE_OBJC_EXCEPTIONS = YES;
-				GCC_PREPROCESSOR_DEFINITIONS = (
-					"DEBUG=1",
-					"$(inherited)",
-				);
```
jeffctown: These were added at the target level, and caused `DEBUG=1` to end up in the resolved macros twice.

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1572,10 +1568,6 @@
 				GCC_ENABLE_OBJC_EXCEPTIONS = YES;
 				GCC_PRECOMPILE_PREFIX_HEADER = YES;
 				GCC_PREFIX_HEADER = "UnitTests/UnitTests-Prefix.pch";
-				GCC_PREPROCESSOR_DEFINITIONS = (
-					"DEBUG=1",
-					"$(inherited)",
-				);
```
jeffctown: These were added at the target level, and caused `DEBUG=1` to end up in the resolved macros twice.

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1623,10 +1615,6 @@
 				DYLIB_CURRENT_VERSION = "$(CURRENT_PROJECT_VERSION)";
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
-				GCC_PREPROCESSOR_DEFINITIONS = (
-					"DEBUG=1",
-					"$(inherited)",
-				);
```
jeffctown: These were added at the target level, and caused `DEBUG=1` to end up in the resolved macros twice.

OHHTTPStubs/OHHTTPStubs.xcodeproj/project.pbxproj
```diff
@@ -1707,10 +1695,6 @@
...
```
jeffctown: These were added at the target level, and caused `DEBUG=1` to end up in the resolved macros twice.

.travis.yml
```diff
@@ -10,17 +10,17 @@ before_install:
 matrix:
   include:
     - osx_image: xcode8.3
-      env: RAKETASK="ios[iOS Framework,latest,build test,-DOHHTTPSTUBS_SKIP_TIMING_TESTS=1]"
+      env: RAKETASK="ios[iOS Framework,latest,build test,OHHTTPSTUBS_SKIP_TIMING_TESTS=1 OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1]"
     - osx_image: xcode8.3
-      env: RAKETASK="osx[Mac Framework,x86_64,build test,-DOHHTTPSTUBS_SKIP_TIMING_TESTS=1]"
+      env: RAKETASK="osx[Mac Framework,x86_64,build test,OHHTTPSTUBS_SKIP_TIMING_TESTS=1 OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1]"
```
AliSoftware: I'm happy surprised that the space character added in there doesn't break anything btw 😅 I guess I properly remembered in the CI config file to use quotes around `"$RAKETASK"` when incoming take, pfew 😄

AliSoftware:
Man it's so refreshing to see CI turn green on the PR without having to restart every CI job manually and pray until it passes 😄🎉

PR #306, opened 2019-04-05
Update Swift 5

CRoig:
### Checklist

### Description

I have update Swift default version to 5.0

### Motivation and Context

Updating project to Swift 5.0 removes a Carthage warning when updating packages using XCode 10.2.
```
***  Skipped installing OHHTTPStubs.framework binary due to the error:
	"Incompatible Swift version - framework was built with 4.2.1 (swiftlang-1000.11.42 clang-1000.11.45.1) and the local version is 5.0 (swiftlang-1001.0.69.5 clang-1001.0.46.3).
```

I built OHHTPPStubs using Carthage using Swift 5.0. Warning is not longer appearing. Also I verified the changes did not break any existing unit test.

### DIFF
--- a/OHHTTPStubs.podspec
+++ b/OHHTTPStubs.podspec
@@ -33,8 +33,6 @@ Pod::Spec.new do |s|
   s.watchos.deployment_target = '2.0'
   s.tvos.deployment_target = '9.0'
 
-  s.swift_version = '5.0'
-
   s.default_subspec = 'Default'
   # Default subspec that includes the most commonly-used components
   s.subspec 'Default' do |default|
--- a/OHHTTPStubs.podspec
+++ b/OHHTTPStubs.podspec
@@ -33,8 +33,9 @@ Pod::Spec.new do |s|
   s.watchos.deployment_target = '2.0'
   s.tvos.deployment_target = '9.0'
 
-  s.default_subspec = 'Default'
+  s.swift_version = '5.0'
 
+  s.default_subspec = 'Default'
   # Default subspec that includes the most commonly-used components
   s.subspec 'Default' do |default|
     default.dependency 'OHHTTPStubs/Core'
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,14 @@ matrix:
       env: RAKETASK="tvos[tvOS Framework,latest,build-for-testing test-without-building,OHHTTPSTUBS_SKIP_TIMING_TESTS=1 OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1]"
     - osx_image: xcode10.1
       env: RAKETASK="osx[Mac Framework,x86_64,build-for-testing test-without-building,OHHTTPSTUBS_SKIP_TIMING_TESTS=1 OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1]"
+    - osx_image: xcode10.2
+      env: RAKETASK="ios[iOS StaticLib,latest,build-for-testing test-without-building,OHHTTPSTUBS_SKIP_TIMING_TESTS=1 OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1]"
+    - osx_image: xcode10.2
+      env: RAKETASK="ios[iOS Framework,latest,build-for-testing test-without-building,OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1]"
+    - osx_image: xcode10.2
+      env: RAKETASK="tvos[tvOS Framework,latest,build-for-testing test-without-building,OHHTTPSTUBS_SKIP_TIMING_TESTS=1 OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1]"
+    - osx_image: xcode10.2
+      env: RAKETASK="osx[Mac Framework,x86_64,build-for-testing test-without-building,OHHTTPSTUBS_SKIP_TIMING_TESTS=1 OHHTTPSTUBS_SKIP_REDIRECT_TESTS=1]"
 
 
 script:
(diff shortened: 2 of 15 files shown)


CRoig:
Updated README with Swift5 compatibility and added XCode 10.2 image to the rakefile. No change required on podspec since XCode 10.2 still supports 8.0 iOS.

AliSoftware:
Thanks!

I just realized that indeed for OHHTTPStubs we haven't updated the podspec with the latest recommendations by CocoaPods as I thought we did. That's is, now CocoaPods supports a `swift_version` attribute that you can add to your podspec to indicate which version(s) of Swift is supported by the pod. In our case that would mean add this attribute to the 'Swift' subspec then.

That `swift_version` has been [introduced in CocoaPods 1.4](http://blog.cocoapods.org/CocoaPods-1.4.0/#swift-version-dsl), and [in CocoaPods 1.7 they introduced](http://blog.cocoapods.org/CocoaPods-1.7.0-beta/) support for a _range_ of Swift versions as well using `swift_versions`. So that migration of OHHTTPStubs to Swift 5 might be the occasion to take advantage of that in our podspec 😉

AliSoftware:
Mmmh having `SWIFT_VERSION` set to 5.0 in the project prevents it to be built using older Xcodes, which is why the CI fails. We need to address that before considering merging

jeffctown:
@AliSoftware Would you be cool using `swift_version` for this pull request, considering that Cocoapods 1.7.0 is still in beta?  I think it makes sense to update to the ranged version of this parameter (`swift_versions`) after the new Cocoapods is fully released.

I like your suggestion to add `swift_version` to the `Swift` subspec, but don't think you can add that parameter to a subspec, or at least thats what Cocoapods complained about when I tried to add it there.  I saw this error (using Cocoapods v1.6.1):

Screen Shot 2019-04-09 at 7 33 59 PM

As for getting CI to pass: I think we can just add the swift version as a `rake` parameter to override the project setting being set to 5.0.  [I have a branch](https://travis-ci.org/AliSoftware/OHHTTPStubs/builds/518042381?utm_source=github_status&utm_medium=notification) that is passing using that method.

@CRoig would you be okay to grant me access to your fork?  I'd like to update some of the project settings the swift migration added before we merge this in.  I can also add the changes to get the CI passing.

AliSoftware:
Yep looks like a good compromise indeed 👍

CRoig:
@jeffctown No problem at all. I did not have the time yet to fix this. But feel free to fork and amend the PR.

jeffctown:
@AliSoftware looks good to me.  Let me know what you think.

OHHTTPStubs.podspec
```diff
@@ -33,8 +33,9 @@ Pod::Spec.new do |s|
   s.watchos.deployment_target = '2.0'
   s.tvos.deployment_target = '9.0'
 
-  s.default_subspec = 'Default'
+  s.swift_version = '5.0'
```
AliSoftware: Just realized, that might mean that CocoaPods will refuse to install the pod in a 4.x project even though it's compatible…

Maybe until we use the range attribute when next CP is out of beta, we should use a lower value here?

Worth manually checking with a project still using Xcode 9 or 10.0 and Swift 4 at least

jeffctown: I think you're right that this parameter is going to cause problems.

## Setting `swift_version` to 5.0

When the `.podspec` has `swift_version` set to 5.0 and I have set the Swift Example project to use Swift 4.0, I can't compile the project with Xcode 10.1.

Screen Shot 2019-04-10 at 8 40 26 PM

A `pod install` or `pod update` will update the `Pods` project to use Swift 4.2, which is then overridden by the OHHTTPStubs target.  

Screen Shot 2019-04-10 at 8 44 30 PM

This could be fixed in Swift 4.0 projects by requiring a `post_install` step in Podfiles, something along the lines of 

```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    if ['OHHTTPStubs'].include? target.name
      target.build_configurations.each do |config|
        config.build_settings['SWIFT_VERSION'] = '4.0'
      end
    end
  end
end
```

We could update the README with this, but it would just add extra work to anyone updating or installing the project, and could result in frustration.  

### Removing `swift_version`

If we remove `swift_version` from the `.podspec`, Cocoapods will use whatever swift version the compiled app is set to use.

Here's Xcode 10.1 and Swift 4, after `pod install`:

Screen Shot 2019-04-10 at 9 00 22 PM

And here's Xcode 10.2 and Swift 5 after `pod install`:

Screen Shot 2019-04-10 at 9 02 55 PM

 

After all of this, I think we should remove `swift_version` from the `.podspec`, do a release for Swift 5.0, then we can add the range in the next release when CocoaPods 1.7.0 is fully released.

AliSoftware: What a detailed investigation 😅
Yes I agree until CP 1.7 is out and the attribute accepting a range of versions is supported, we should remove the swift_version attribute from the podspec then.

eMdOS: Any plan for merging this soon? Seems like it is ready to be deployed 😬.
My project is fully migrated to Swift 5, just waiting for this release in order to re-enable my network stubs for unit testing purposes.

AliSoftware: @eMdOS as you can see from the conversation above, merging this as-is would mean breaking the backward compatibility with older Xcode and study versions, so we need to address that first by fixing the podspec, only then child we release 😅

jeffctown: I think this one's ready to roll now @AliSoftware .

jeffctown: @Alisoftware - friendly bump. :)

AliSoftware:
🎉

jeffctown:
@CRoig - I just finished the 8.0.0 release for Swift 5.  Thanks for the help!

@eMdOS - FYI

PR #308, opened 2019-04-22
Release Prep for 8.0.0

jeffctown:
### Checklist

- [X] I've updated the documentation if necessary

### Description

Release Preparation for Swift 5.0

### Motivation and Context

- Bumped version to 8.0.0.
- Updated CHANGELOG.md with release version.
- Updated example projects to use 8.0.0 pod.
- Updated both examples projects using Xcode 10.2 to silence all warnings.
- Pod lib lint passing:
Screen Shot 2019-04-22 at 8 20 05 AM


jeffctown:
Thats a good question @AliSoftware .  I think it probably depends on what happens with older versions of cocoa pods when you add that parameter to a pod file.  Will it require users to update Cocoapods or not?

AliSoftware:
@jeffctown Wait a minute I might have approved that a bit quickly: the API of the lib didn't change, in fact there was no change need to sort Swift 5… so should we really make this version a "breaking change" by bumping to 8.0.0 already?

jeffctown:
@AliSoftware uh oh!  My apologies, I thought that was expected since that was what we did for supporting Swift 4 too.  What do you think is the right action to take now?

AliSoftware:
Well since you already did the release and pushed it to trunk, I guess it's too late now 😉 

Don't worry, that's not such a big deal. It's just that people might be more reluctant to migrate to a major version bump (as according to semantic versioning that is supposed to mean "the API changed and some old methods don't exist anymore so if you upgrade you'll need to modify your calling code in your app", as opposed to minor version bump meaning "there are new methods exposed in the API but the old ones are still present so existing code will still work, it's just additive" and patch bump meaning "the API didn't change at all we just fixed some bugs in the implementations").

So people might fear that migrating to 8.0.0 will require some work and refactoring in their calling code to OHHTTPStubs while it actually doesn't require any change... but it's less problematic than if we made the mistake in the other direction 😜

PR #315, opened 2019-09-18
Better nullability and error handling

### LINKED ISSUE #290: Static Analysis results ran from XCode 9.4.1
Ran Static analysis on my project and saw some collateral information that I thought I'd pass on.
I realize this is older code but probably still applies.

- OHHTTPStubs/OHHTTPStubs/Sources/OHHTTPStubs.m:46:1: Property of mutable type 'NSMutableArray' has 'copy' attribute; an immutable object will be stored instead

**Memory error Group**

- OHHTTPStubs/OHHTTPStubs/Sources/Mocktail/OHHTTPStubs+Mocktail.m:44:9: nil returned from a method that is expected to return a non-null value

- OHHTTPStubs/OHHTTPStubs/Sources/Mocktail/OHHTTPStubs+Mocktail.m:103:9: nil returned from a method that is expected to return a non-null value

- OHHTTPStubs/OHHTTPStubs/Sources/OHHTTPStubsResponse.m:153:12: nil passed to a callee that requires a non-null 1st parameter

### New Issue Checklist

- [NO ] I have tried with the latest version of OHHTTPStubs
- [YES ] I have read the [README](https://github.com/AliSoftware/OHHTTPStubs/blob/master/README.md)
- [ N/A] I have read the [Using the right Swift Version of `OHHTTPStubs` for your project](https://github.com/AliSoftware/OHHTTPStubs#using-the-right-swift-version-of-ohhttpstubs-for-your-project) section
- [NO ] I have searched in the [existing issues](https://github.com/AliSoftware/OHHTTPStubs/issues?utf8=✓&q=is%3Aissue)
- [ N/A] I have read [the OHHTTPStubs wiki](https://github.com/AliSoftware/OHHTTPStubs/wiki) to see if there wasn't a detailed page talking about my issue

### Environment

- version of OHHTTPStubs: [LIB VERSION HERE]
- integration method you are using:
  * [ ] Cocoapods
  * [ ] Carthage
  * [ ] submodule
  * [X ] other (*Built into our Project from sources)
- version of the tool you use: [5.1.0]

There is no ill issue seen with OHTTPStubs. This is just FYI in case you wish to investigate. 
I realize this is an older version. I can't update and try again.

ZevEisenberg:
### Checklist

### Description

Fixes #290

Also does more idiomatic error handling:

#### Wrong:

```objc
NSError *error;
NSArray *result = [someObject someMethod error:&error];
if (!error) {
    // assume success
}
```

#### Correct:
```objc
NSError *error;
NSArray *result = [someObject someMethod error:&error];
if (result) {
    // assume success
}
```

### Motivation and Context

The main motivation was addressing static analysis warnings I encountered when analyzing an Objective-C project. The error handling was just something I fixed along the way.

Note: I also added a couple of missing generics annotations which may affect the public interface. If they are imported into Swift, they will go from `[Any]` to `[OHHTTPStubsDescriptor]`. This is a breaking change for anyone who is using the output of this method in Swift.

Testing: I added a couple of missing tests for error throwing cases, and made sure all the tests were passing.

### DIFF
--- a/OHHTTPStubs/Sources/OHHTTPStubs.m
+++ b/OHHTTPStubs/Sources/OHHTTPStubs.m
@@ -43,7 +43,7 @@ static NSTimeInterval const kSlotTime = 0.25; // Must be >0. We will send a chun
 
 @interface OHHTTPStubs()
 + (instancetype)sharedInstance;
-@property(atomic, copy) NSMutableArray* stubDescriptors;
+@property(atomic, readonly) NSMutableArray* stubDescriptors;
 @property(atomic, assign) BOOL enabledState;
 @property(atomic, copy, nullable) void (^onStubActivationBlock)(NSURLRequest*, id<OHHTTPStubsDescriptor>, OHHTTPStubsResponse*);
 @property(atomic, copy, nullable) void (^onStubRedirectBlock)(NSURLRequest*, NSURLRequest*, id<OHHTTPStubsDescriptor>, OHHTTPStubsResponse*);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # OHHTTPStubs — CHANGELOG
 
+<!-- Add new release notes above the latest release. The new version number will get added last before the new release. -->
+
 ## [8.0.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/8.0.0)
 
 * Update default Swift Version to 5.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 [@croig](https://github.com/CRoig)
 
 >Notes:
-> * No code changes were required (except from a little missing comma which caused a compilation error). Only xcshemes and xcodeproj were changed.
+> * No code changes were required (except from a little missing comma which caused a compilation error). Only xcschemes and xcodeproj were changed.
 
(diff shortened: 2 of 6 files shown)


OHHTTPStubs/Sources/OHHTTPStubs.m
```diff
@@ -43,7 +43,7 @@ @interface OHHTTPStubsProtocol : NSURLProtocol @end
 
 @interface OHHTTPStubs()
 + (instancetype)sharedInstance;
-@property(atomic, copy) NSMutableArray* stubDescriptors;
+@property(atomic, readonly) NSMutableArray* stubDescriptors;
```
AliSoftware: Problem with this is, it's a reference type. So that means that users can't set `stubDescriptors = newValue` but they can `[stubDescriptors append:…]` though can't they?

AliSoftware: Oh wait it's in the private category so not exposed to the consumer is it? (Not easy to review from phone 😅)

ZevEisenberg: (Forgot to reply originally) `copy` never works on properties of `NSMutableCopying` type, because the setter calls `-copy`, which returns an immutable copy, i.e. an `NSArray` in this case. So you’d set a mutable array, and the value that gets stored would be immutable. Fun edge case in Obj-C land that should probably be a compile error.

PR #319, opened 2019-10-23
Add missing import to support building with Bazel

thii:
### Checklist

### Description
Add missing import to support building with Bazel, since it doesn't import underlying system modules by default.


jeffctown:
@thii Hey!  Thanks so much for putting this up.  If you edit the changelog for this PR, I don't see a problem merging it in.  

Do you by any chance have the time to add a BUILD file so other Bazel users can integrate with this repo easily?

thii:
@jeffctown Unfortunately we can't open source the BUILD file for it now.

This only adds a missing import (which implicitly gets imported if building with Xcode) so I think it's OK to skip the changelog to keep it clean.

thii:
I've updated the changelog and added a BUILD file for it.

jeffctown (approved):

thii:
Looks like the BUILD file was moved to a different directory, probably unintentionally?

jeffctown:
@thii good catch.  I'm not sure how that happened, but I'll clean it up this weekend and do a v10 release.

PR #337, opened 2020-12-01
Updating for 9.1.0 release.

jeffctown:
Updating version in Changelog.
Updating Example projects.
Updating podspec.

### Checklist

- [X] I've updated the documentation if necessary

### Description

Updating for release 9.1.0.


Examples/Swift/Podfile.lock
```diff
@@ -24,8 +24,8 @@ EXTERNAL SOURCES:
     :path: "../.."
 
 SPEC CHECKSUMS:
-  OHHTTPStubs: cb29d2a9d09a828ecb93349a2b0c64f99e0db89f
+  OHHTTPStubs: 90eac6d8f2c18317baeca36698523dc67c513831
 
 PODFILE CHECKSUM: 7da7c441ea9ff6f06b633c908b7a7294805f5602
 
-COCOAPODS: 1.7.5
+COCOAPODS: 1.9.3
```
Kaspik, line 31: 1.9.3 is not compatible with Ruby 2.7.0, is there a reason why not to use 1.10.0? :)

Kaspik (approved):
Thanks!

AliSoftware (approved):
LGTM!

Examples/Swift/Podfile.lock
AliSoftware: That's a good question. Not sure it matters here because it's on the `Podfile.lock` of the example projects (and it shouldn't prevent users who use CO 10.0 to `pod install` on the example projects if they wanted to anyway) so it's not a restriction on the pod itself or its podspec. But good to keep in mind (I didn't know about the ruby versions compatibility limitation with 1.9.3)

Kaspik: ohh, right, example - yeah it doesn't then :)

jeffctown:
I [tagged this](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/9.1.0) for you @Kaspik .  Thanks for reminding me!

Kaspik:
Thank youuu, last XCode 12 warning is fixed now! 🎉

AliSoftware:
Thanks for taking care of this @jeffctown ! 🎉

PR #367, opened 2024-03-07
Specify ruby version and use bundler to fix gem versions

Liquidsoul:
### Checklist

### Description

- add a `.ruby-version` to specify the ruby version to use.
- add a `Gemfile` to specify the version of the gem to use, mostly for `rake` and `cocoapods`.

### Motivation and Context

When developing on the project, it can be difficult to ensure a consistent environment when the tools used or their versions are not specified somewhere.  
This PR tries to help by specifying the ruby environment that should be used when running the various tasks.


AliSoftware:
Done some similar work in https://github.com/AliSoftware/OHHTTPStubs/pull/368 just now

PR #377, opened 2025-03-26
Exposed removeAllStubs from Swift API

AllDmeat:
### Checklist

### Description

In my tests I'm importing `OHTTPStubsSwift` for stubbing, but I need to import `OHTTPStubs` as well just for stubs removal.
I've added `removeAllStubs()` method right beside other Swift Syntax Sugar helpers to omit this requirement.


AllDmeat:
@AliSoftware

